PR #30, opened 2019-01-22
Add wheel in the venv

subos2008:


iamdefinitelyahuman:
Install wheel in venv

PR #35, opened 2019-02-07
Add Contract.methods()

subos2008:
Facilitates the whitelisting of public ABI methods in tests


iamdefinitelyahuman:
This is already available as Contract.signatures

PR #43, opened 2019-02-24
Add sudo to Dockerfile

sot528:
I could not build this Docker image. So I added `sudo` to Dockerfile.


iamdefinitelyahuman:
Thanks :)

PR #50, opened 2019-03-01
Update docker tag

nymd:
If you have "brownie:1" in the build command, the image that results is tagged as brownie:1.

The run command doesn't specify brownie:1 so docker looks for brownie:latest, which doesn't exist.  This PR removes the 1. Alternately you could add the :1 to the run command.


subos2008:
Looks good to me

PR #170, opened 2019-08-25
Add PEP484 Annotations

crawfordleeds:
### What was wrong / missing?

This is a WIP PR to add PEP484 annotations to the codebase and update the docs accordingly.

### How was it fixed / added?

Will report back soon.


iamdefinitelyahuman:
It seems the annotations are messing with the delicate import balance, resulting in circular import issues.

The heart of the problem is occuring in `brownie.network.__init__`:

https://github.com/iamdefinitelyahuman/brownie/blob/493f0f6d1202754f5adf7c72d5251f9c2ef2aadb/brownie/network/__init__.py#L11-L22

`TxHistory` is being instantiated as `history`, which creates a potential namspace collision when attempting to import `brownie.network.history` as a package.  There is a similar situation for `rpc` and `web3`, however the only thing inside each of these packages is the singleton class so there's never a need to import the packages when the instantiated object is directly available.

Possible solutions I see:

1. Stop instantiating `TxHistory` as `history` in `brownie.network.__init__` - this is the most obvious choice, but when i started to do it I realized that _everything_ is importing `brownie.network.history` and expecting the singleton object. Changing this would lead to changes all over the project.

2. Rename `brownie.network.history` - along with the `TxHistory` singleton, this module also acts as a single point to track all deployed contracts. Other possible names that i think would fit are `state` or `chain`. This change would touch a few files, but be far less messy than the previous suggestion.

3. In `brownie.network.contract`, change `from . import history` to:
    ```python
    from .history import (
      _add_contract,
      _remove_contract,
      find_contract,
    )
    ```
    Modifying the import statement in this way manages to circumvent the name collision. It's the simplest fix by far, but it feels kinda quick and dirty to me and like exactly the sort of thing that i'll hate myself for in 3 months when something else related pops up.

I'm inclined towards number 2, but not fully opposed to 3 and open to other suggestions.

crawfordleeds:
Hey @iamdefinitelyahuman, I'm down to just one error on the build in `tests/test/plugin/test_session_fixtures.py`. My intuition says it has something to do with the difference between `history` as the module and `history` as an instance of TxHistory that is getting confused.

I'm going to be offline for a few hours so I thought I'd ask if the issue seems obvious to you?

Further, I think some refactoring of the tests may be in order for the various tests around `history`.

brownie/__init__.py
```diff
@@ -19,7 +19,7 @@
 __all__ = [
     'accounts',
     'alert',
-    'history',
+    'state',
```
iamdefinitelyahuman: This is a great example of why going with option 2 instead of 3 is a better choice.  In this case, you do still want to import the instantiated version of `TxHistory` as `history`.  If you reverse this change, that last test should be passing again.

brownie/network/account.py
```diff
@@ -259,7 +275,13 @@ def estimate_gas(self, to, amount, data=""):
...
             raise
 
-    def transfer(self, to, amount, gas_limit=None, gas_price=None, data=""):
+    def transfer(
+            self,
+            to: 'Accounts',
+            amount: int,
+            gas_limit: float = None,
+            gas_price: float = None,
+            data: str = "") -> 'TransactionReceipt':
```
iamdefinitelyahuman: this looks so much cleaner than putting the `):` on the next line like a lonely frowny face :+1:

brownie/network/account.py
```diff
@@ -388,13 +410,13 @@ def __eq__(self, other):
...
 
-    def balance(self):
+    def balance(self) -> Wei:
         '''Returns the current balance at the address, in wei.'''
         balance = web3.eth.getBalance(self.address)
         return Wei(balance)
 
 
-def _raise_or_return_tx(exc):
+def _raise_or_return_tx(exc: Any) -> Any:
```
iamdefinitelyahuman: pretty sure `exc` will only ever be a `ValueError`

brownie/network/contract.py
```diff
@@ -55,59 +63,63 @@ class ContractContainer(_ContractBase):
...
             len(web3.eth.getCode(i.address).hex()) <= 4
         ]
         for contract in reverted:
-            history._remove_contract(contract)
+            _remove_contract(contract)
             self._contracts.remove(contract)
             contract._reverted = True
 
-    def remove(self, contract):
+    def remove(self, contract: str) -> None:
```
iamdefinitelyahuman: `contract` will be a `Contract` instance, but maybe this is to avoid some mypy confusion?

brownie/network/contract.py
```diff
@@ -136,7 +148,7 @@ class ContractConstructor:
 
     _dir_color = "contract_method"
 
-    def __init__(self, parent, name):
+    def __init__(self, parent: Any, name: str) -> None:
```
iamdefinitelyahuman: `parent` is a `Contract` instance

brownie/network/contract.py
```diff
@@ -250,23 +262,23 @@ def __eq__(self, other):
...
         return super().__eq__(other)
 
-    def __getattribute__(self, name):
+    def __getattribute__(self, name: str) -> Any:
         if super().__getattribute__('_reverted'):
             raise ContractNotFound("This contract no longer exists.")
         return super().__getattribute__(name)
 
-    def balance(self):
+    def balance(self) -> int:
```
iamdefinitelyahuman: `Wei`

brownie/network/contract.py
```diff
@@ -462,13 +482,17 @@ def _get_tx(owner, args):
     return args, tx
 
 
-def _get_method_object(address, abi, name, owner):
+def _get_method_object(
+        address: str,
+        abi: Any,
```
iamdefinitelyahuman: `abi` should always be a `Dict`

brownie/network/contract.py
```diff
@@ -462,13 +482,17 @@ def _get_tx(owner, args):
...
+        name: str,
+        owner: Optional[AccountsType]) -> Union['ContractCall', 'ContractTx']:
     if abi['stateMutability'] in ('view', 'pure'):
         return ContractCall(address, abi, name, owner)
     return ContractTx(address, abi, name, owner)
 
 
-def _params(abi_params):
+def _params(abi_params: List) -> Any:
```
iamdefinitelyahuman: always returns a `List`

brownie/network/contract.py
```diff
@@ -478,12 +502,12 @@ def _params(abi_params):
     return types
 
 
-def _inputs(abi):
+def _inputs(abi: Any) -> str:
```
iamdefinitelyahuman: here and line 510, `abi` is `Dict`.  Guessing this is again something to do with mypy confusion?

brownie/network/event.py
```diff
@@ -172,15 +177,15 @@ def get_topics(abi):
     return eth_event.get_topics(abi)
 
 
-def decode_logs(logs):
+def decode_logs(logs: Optional[str]) -> Union['EventDict', List[None]]:
```
iamdefinitelyahuman: `logs` will be a `List`..  I don't think it's optional either, the check for truthyness is just to see if the list is empty.

brownie/network/event.py
```diff
@@ -172,15 +177,15 @@ def get_topics(abi):
...
     if not logs:
         return []
     events = eth_event.decode_logs(logs, _topics)
     events = [format_event(i) for i in events]
     return EventDict(events)
 
 
-def decode_trace(trace):
+def decode_trace(trace: Optional[str]) -> Union['EventDict', List[None]]:
```
iamdefinitelyahuman: Same as above..  `List` and not optional.

brownie/network/transaction.py
```diff
@@ -117,6 +126,8 @@ def __init__(self, txid, sender=None, silent=False, name='', revert_data=None):
         self.status = -1
         self.txid = txid
         self.fn_name = name
+        self._revert_pc: Any
```
iamdefinitelyahuman: What's happening here?

brownie/network/transaction.py
```diff
@@ -158,14 +169,14 @@ def __init__(self, txid, sender=None, silent=False, name='', revert_data=None):
...
+    def __repr__(self) -> str:
         c = {-1: 'pending', 0: 'error', 1: None}
         return f"<Transaction object '{color[c[self.status]]}{self.txid}{color}'>"
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(self.txid)
 
-    def __getattr__(self, attr):
+    def __getattr__(self, attr: Any) -> None:
```
iamdefinitelyahuman: This should probably return `Any`

brownie/network/transaction.py
```diff
@@ -630,7 +641,13 @@ def _source_string(self, idx, pad):
         )
 
 
-def _format_source(source, linenos, path, pc, idx, fn_name):
+def _format_source(
+        source: Any,  # source: 'Accounts'
```
iamdefinitelyahuman: Avoiding dependency issues?

brownie/project/compiler.py
```diff
@@ -505,12 +528,16 @@ def expand_source_map(source_map):
     return source_map
 
 
-def _expand_row(row):
-    row = row.split(':')
-    return [int(i) if i else None for i in row[:3]] + row[3:] + [None] * (4 - len(row))
+def _expand_row(row_str: str) -> Any:
+    row = row_str.split(':')
+    # ignore typing of next line
```
iamdefinitelyahuman: Agreed, this is ugly.  I had to generate similar code for Vyper's source map and it looks a lot cleaner: https://github.com/ethereum/vyper/blob/faacf1503e02b7af935a6dcf8eda4ec697ed2101/vyper/compiler.py#L177-L196

Feel free to remove this comment and raise another issue about how ugly my code golf looks :rofl:

brownie/project/scripts.py
```diff
@@ -87,14 +95,16 @@ def _get_path(path_str, default_folder="scripts"):
     return path
 
 
-def _import_from_path(path):
+# TODO: Come back to this
+# Need to check if there are any cases where the argument path may be a str instead of Path
+def _import_from_path(path: Any) -> ModuleType:
```
iamdefinitelyahuman: It will only ever be `Path`

brownie/project/sources.py
```diff
@@ -108,7 +109,7 @@ def is_inside_offset(inner, outer):
     return outer[0] <= inner[0] <= inner[1] <= outer[1]
 
 
-def get_hash(source, contract_name, minified):
+def get_hash(source: Any, contract_name: str, minified: bool) -> str:
```
iamdefinitelyahuman: `source` is always `str` i believe

brownie/project/sources.py
```diff
@@ -120,7 +121,8 @@ def get_hash(source, contract_name, minified):
         return ""
 
 
-def highlight_source(source, offset, pad=3):
+def highlight_source(source: Any, offset: Tuple, pad: int = 3) -> Any:
+    # TODO: Reconcile the args in the docs below, and return type
```
iamdefinitelyahuman: I think we resolved this, right?

docs/api-brownie.rst
```diff
@@ -27,14 +27,16 @@ Formatting Contract Data
...
 
     Formats inputs based on a contract method ABI.
 
+    Returns
+
     * ``abi``: A contract method ABI as a dict.
     * ``inputs``: List or tuple of values to format.
 
-    Returns a list of values formatted for use by ``ContractTx`` or ``ContractCall``.
+    Returns a tuple subclass (brownie.convert.ReturnValue) of values formatted for use by ``ContractTx`` or ``ContractCall``.
```
iamdefinitelyahuman: nice catch

docs/api-brownie.rst
```diff
@@ -43,12 +45,14 @@ The following methods are used to convert multiple values based on a contract AB
         >>> from brownie.convert import format_input
         >>> abi = {'constant': False, 'inputs': [{'name': '_to', 'type': 'address'}, {'name': '_value', 'type': 'uint256'}], 'name': 'transfer', 'outputs': [{'name': '', 'type': 'bool'}], 'payable': False, 'stateMutability': 'nonpayable', 'type': 'function'}
         >>> format_input(abi, ["0xB8c77482e45F1F44dE1745F52C74426C631bDD52","1 ether"])
-        ['0xB8c77482e45F1F44dE1745F52C74426C631bDD52', 1000000000000000000]
+        ('0xB8c77482e45F1F44dE1745F52C74426C631bDD52', 0)
```
iamdefinitelyahuman: Why did you make this change? It seems incorrect to me.

docs/api-brownie.rst
```diff
@@ -127,7 +131,7 @@ The following classes and methods are used to convert arguments supplied to ``Co
         0xFF
         >>> bytes_to_hex("Hello")
           File "brownie/types/convert.py", line 149, in bytes_to_hex
-            raise ValueError("'{}' is not a valid hex string".format(value))
+            raise ValueError("'{value}' is not a valid hex string".format(value))
```
iamdefinitelyahuman: `f"'{value}' is not a valid hex string"`  - good eye noticing this was outdated tho

brownie/network/account.py
crawfordleeds: PEP8 FTW!

docs/api-brownie.rst
crawfordleeds: Ah yes my bad! This was when I broke the _to_wei and forgot to fix it. Thanks!

brownie/_config.py
```diff
@@ -15,43 +16,43 @@
...
-    def __setitem__(self, key, value):
+    def __setitem__(self, key: Any, value: Any) -> None:
         if self._locked and key not in self:
             raise KeyError(f"{key} is not a known config setting")
         if type(value) is dict:
             value = ConfigDict(value)
         super().__setitem__(key, value)
 
-    def update(self, arg):
+    def update(self, arg):  # type: ignore
```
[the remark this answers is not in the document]
crawfordleeds: That's what I thought but I get an odd error that I haven't seen before. As soon I remove `# type: ignore` I get the following error when running mypy:

```
brownie/_config.py:31: error: Signature of "update" incompatible with supertype "dict"
brownie/_config.py:31: error: Signature of "update" incompatible with supertype "MutableMapping"
```

Have you seen that before?

Note this occurs whether or not I declare arg as a `Dict` or `Any`.

brownie/convert.py
```diff
@@ -32,92 +35,92 @@ class Wei(int):
...
         return Wei(super().__sub__(_to_wei(other)))
 
 
-def _to_wei(value):
+def _to_wei(value: WeiInputTypes) -> int:
     original = value
     if value is None:
         return 0
-    if type(value) in (bytes, HexBytes):
+    if isinstance(value, bytes) or isinstance(value, HexBytes):
```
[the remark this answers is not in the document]
crawfordleeds: Good catch, thanks!

brownie/network/account.py
```diff
@@ -28,47 +30,47 @@ class Accounts(metaclass=_Singleton):
...
-    def _reset(self):
+    def _reset(self) -> None:
         self._accounts.clear()
         try:
             self._accounts = [Account(i) for i in web3.eth.accounts]
         except Exception:
             pass
 
-    def _revert(self, height):
+    def _revert(self, height: Any) -> None:
```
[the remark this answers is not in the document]
crawfordleeds: That seems right. However, `height` isn't actually used in this method. Is this a relic of some previous functionality? Probably makes sense to open as a separate issue, let's just get this merged as it :)

brownie/_config.py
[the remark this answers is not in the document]
iamdefinitelyahuman: That's odd.  Anyway it's within a private area of the API and it's a pretty common method on a dict type, so I'm sure it's fine to leave.

brownie/network/account.py
crawfordleeds: That seems to check out. Done.

brownie/network/account.py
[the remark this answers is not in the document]
iamdefinitelyahuman: Related - #194

Issue #197, opened 2019-09-12
Rename `history` tests as `state` where appropriate


crawfordleeds:
There are a few tests that are named after the `history` module, which after merging #170, will be called `state`. The tests corresponding to the new `state` module (not `history` as an instance of `TxHistory()` should be renamed appropriately.

PR #206, opened 2019-09-15
Rename state tests

crawfordleeds:
Per #197, renaming tests using the `state` module appropriately as opposed to `history`.


crawfordleeds:
@iamdefinitelyahuman am I missing something or is this as simple as just renaming that one file? I can't find any other references in the tests to `history` that don't mean `TxHistory()`.

PR #207, opened 2019-09-15
Docker container for tests

### LINKED ISSUE #186: Add Docker support for tests
I'm having issues running the tests locally. Since there are a lot of dependencies it would be nice to have a docker image available to quickly run all the tests.

I'll get the issues worked out on my local machine, but I figure if I'm having the issue others will too in the future.

crawfordleeds:
This PR resolves #186 to provide a Docker environment for testing as well as some instructions on how to use it in the README.


README.md
```diff
@@ -50,6 +50,33 @@ $ pip install -r requirements-dev.txt
 
 Then use ``tox`` to run the complete suite against the full set of build targets, or ``pytest`` to run tests against a specific version of Python. If you are using ``pytest`` you must include the ``-p no:pytest-brownie`` flag to prevent it from loading the Brownie plugin.
 
+### Using Docker
+
+You can use a sandbox container provided in the docker-compose.yml file for testing inside a Docker environment.repository
```
nymd: extra word at the end: repository

README.md
```diff
@@ -50,6 +50,33 @@ $ pip install -r requirements-dev.txt
...
+
+Start the test environment:
+
+```bash
+docker-compose up -d
+```
+
+To open a session to the container:
+```bash
+docker-coampose exec sandbox bash
```
nymd: "coampose" -> compose

README.md
```diff
@@ -50,6 +50,33 @@ $ pip install -r requirements-dev.txt
...
+```
+
+To run arbitrary commands, use the `bash -c` prefix.
+```bash
+docker-compose exec sandbox bash -c ''
+```
+
+For example, to run the tests in `brownie/tests/test_format_input.py`:
+```bash
+docker-compose exec sandbox bash -c 'python -m pytest tests/convert/test_format_input.py'
```
nymd: Why should we suggest using compose when it is one container and will likely always be just one container?  Maybe docker run / docker exec would be more direct?

README.md
crawfordleeds: Thanks, autofill on pycharm.

README.md
crawfordleeds: I was debating the same thing myself.  First, this is how web3.py organizes it so it seemed reasonable to provide a similar development experience.

Secondly, maybe this is just my naivety on Docker, but isn't using docker-compose one less command for the user? Here it's just `docker-compose up -d` then `docker exec`. Otherwise, it would be `docker build`, `docker run`, then `docker exec`? docker-compose also removes the need to remember to bind mount the working directory.

I'm fine removing docker-compose if you don't agree with those tradeoffs - let me know.

nymd: You are correct in that it removes a few steps; I'm fine with leaving it. As a docker power user, you could look at the compose and decide that you just want to build/run/exec it yourself.

iamdefinitelyahuman:
closes #186

README.md
crawfordleeds: Thanks!

PR #214, opened 2019-09-19
updates for py-solc-ast 1.0.0

iamdefinitelyahuman:
Updates based on breaking changes to `py-solc-ast`.


iamdefinitelyahuman:
Note: issues with the build are because `solcast` hadn't fully populated on pypi yet.

PR #244, opened 2019-11-08
Dumping private keys when loading ganache-cli on specificied filename.

manfr3d:
### What was wrong / missing?

Filename parameters to dump private keys on running ganache-cli. Can be useful when connecting to Ganache from an external Dapp.

### How was it fixed / added?

Added in rpc.py

### DIFF
--- a/brownie/network/rpc.py
+++ b/brownie/network/rpc.py
@@ -22,6 +22,7 @@ CLI_FLAGS = {
     "accounts": "--accounts",
     "evm_version": "--hardfork",
     "mnemonic": "--mnemonic",
+    "acctKeys": "--acctKeys",
 }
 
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman:
:+1: thank you for the contribution!  Tests are failing from hidden environment variables, nothing to worry about.

Do you mind also updating the documentation on [configuration files](https://github.com/iamdefinitelyahuman/brownie/blob/master/docs/config.rst) and [the local rpc client](https://github.com/iamdefinitelyahuman/brownie/blob/master/docs/test-rpc.rst)? Then I think this will be ready to merge.

manfr3d:
I update the documentation. Thanks!

Issue #262, opened 2019-11-23
Brownie - rebuild / expand CLI test cases


iamdefinitelyahuman:
### The Issue
1. The test cases in `tests/cli/test_cli_main.py` are poorly designed. Tests are using a custom class `CliTester` which provides a similar functionality to `monkeypatch.setattr` and `mocker.spy`, but is difficult to understand on first glance.
2. Test coverage for some cli modules is lacking, especially `brownie/_cli/analyze.py` and `brownie/_cli/ethpm.py`.

### Completion Requirements
1. Rewrite tests in `tests/cli/test_cli_main.py` to use `monkeypatch` and  `pytest-mock`. The `CliTester` class should no longer exist.
2. Increase test coverage for all files in `brownie/_cli/` to >=90%

### Implementation Notes
Functionality that is exists outside of the `brownie/_cli/` folder is already covered elsewhere in the test suite. To maximize speed of the tests, you should patch these methods. The important thing is to verify that the CLI is correctly calling the intended methods.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.351 ETH (49.98 USD @ $142.39/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/iamdefinitelyahuman/brownie/262/3810).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/iamdefinitelyahuman/brownie/262/3810).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $94,082.76 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


[...]

PR #273, opened 2019-12-15
Replaces CliTester functionalities with monkeypatch and mocker.spy

agbilotia1998:
### What was wrong / missing?
Refers: #262 
test_cli_main.py had CliTester class implemented performing functions similar to monkeypatch.setattr etc.

### How was it fixed / added?
The class functionalities were replaced with `monkeypatch.setattr` and `mocker.spy`.

### DIFF
--- a/tests/cli/test_cli_main.py
+++ b/tests/cli/test_cli_main.py
@@ -5,81 +5,109 @@ import sys
 import pytest
 
 from brownie._cli.__main__ import main as cli_main
+from brownie._cli.console import Console as cli_console
 
 
 @pytest.fixture
-def cli_tester(monkeypatch):
-    c = CliTester(monkeypatch)
-    yield c
-    c.close()
+def cli_tester(monkeypatch, mocker):
+    tester = CliTester(monkeypatch, mocker)
+    tester.mocker.spy(tester, "mock_subroutines")
+
+    yield tester
+    tester.close()
 
 
 class CliTester:
-    def __init__(self, monkeypatch):
+    def __init__(self, monkeypatch, mocker):
         self.argv = sys.argv.copy()
         self.monkeypatch = monkeypatch
-        self.called = False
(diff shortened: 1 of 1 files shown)


agbilotia1998:
@iamdefinitelyahuman please review.  CliTester class is kept as I've monkeypatched the targets and subtargets to a mocked function in this class and as mocker.spy requires an object as the first parameter so the object of this class serves the functionality.

tests/cli/test_cli_main.py
```diff
@@ -5,81 +5,113 @@
...
 
-    def catch(self, *args, **kwargs):
-        assert self.args == args
-        assert self.kwargs == kwargs
-        self.called = True
-
-    def incremental_catch(self, *args, **kwargs):
-        self.count += 1
+    def mock_subroutines(self, *args, **kwargs):
+        print("Monkey Patched Function")
```
iamdefinitelyahuman: rather than printing, i would just `pass` or `return True`

tests/cli/test_cli_main.py
```diff
@@ -5,81 +5,113 @@
...
 
     def close(self):
         sys.argv = self.argv
 
 
 def test_cli_init(cli_tester):
-    cli_tester.set_target("brownie.project.new")
-    cli_tester.run("init", args=(".", False))
-    cli_tester.run("init test/path --force", args=("test/path", True))
+    cli_tester.mocker.spy(cli_tester, "mock_subroutines")
```
iamdefinitelyahuman: this call is repeated in every test - you could incude it at the start of `cli_tester.run_and_test_parameters`

agbilotia1998: Added the statement as a part of the fixture.

iamdefinitelyahuman:
Merging - this completes part 1 of #262, part 2 can be a separate PR

PR #296, opened 2020-01-08
Update ethPM template

fubuloubu:
### What was wrong / missing?
The EthPM template provided with Brownie could be made more explicit with some commenting and re-arranging. Also, several fields were noted as mandatory when they aren't (per the spec).

### How was it fixed / added?
Made the modifications, seeing if the tests work

### DIFF
--- a/brownie/data/ethpm-config.yaml
+++ b/brownie/data/ethpm-config.yaml
@@ -1,27 +1,33 @@
 # Brownie ethPM package configuration file
-# https://eth-brownie.readthedocs.io/en/stable/ethpm.html
+# Read more here: https://eth-brownie.readthedocs.io/en/stable/ethpm.html
 
-# required fields
-package_name:
-version:
-settings:
-  deployment_networks:
-    -
-    -
-  include_dependencies:
+# Required Fields
+package_name: demo-project  # Your project's name goes here (hint: no spaces)
+version: 0.1.0a1  # Your project's version goes here (hint: use semver)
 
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman:
This shouldn't have passed tests, but I don't think there are any test cases to verify the fields of the default `ethpm-config.yaml`.

1. `deployment_networks` can be set to `false` and the generated manifest will not include it. I feel this is an important enough section of a manifest that it should not be possible to implicitly omit it.
2. `include_dependencies` is a brownie setting. Not necessary if a project has no installed dependencies, but IMO much simpler to have the user enter `true`/`false` than raising when attempting to generate a manifest because the setting isn't present.
3. All of the optional fields are omitted if left blank, I only included them as an example of good information to include in `meta`.

My feeling is that what's missing here is adequate documentation.  At minimum the link at the top should point directly to [package config settings](https://eth-brownie.readthedocs.io/en/stable/ethpm.html#step-1-package-configuration-settings). Far better would be comments on each field explaining exactly the purpose and allowed values.

fubuloubu:
I think `include_dependencies` might better live in the brownie config file, or another brownie file that Brownie has control over which acts as a lock file (I guess that's what ethPM packages are supposed to do). Especially when thinking about #299

iamdefinitelyahuman:
Closing this for now. The conversation is covered in several issues and with ethPM v3 on the way, the entire functionality is due for a rework.

PR #311, opened 2020-01-10
feat: Add ETC and Kotti RPC endpoints to config

YazzyYaz:
### What was wrong / missing?
RPC Public endpoints for ETC and Kotti Testnet

### How was it fixed / added?
Add them to config file so it makes the tutorial I'm working on for Brownie easier to do.

### DIFF
--- a/tests/main/test_config.py
+++ b/tests/main/test_config.py
@@ -21,7 +21,7 @@ def test_network_keys(config):
         "rinkeby",
         "ropsten",
         "classic",
-        "kotti"
+        "kotti",
     }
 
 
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman:
Looks good, just need to add the new network names in `tests/main/test_config.py`

Would also be great to add support for `atlantis` and `agartha` as valid evm rulesets in the config file, and then translate them to `byzantium`/`petersburg` prior to launching ganache or solc.  But that's probably worthy of a separate PR.

tests/main/test_config.py
```diff
@@ -20,6 +20,8 @@ def test_network_keys(config):
         "kovan",
         "rinkeby",
         "ropsten",
+        "classic",
+        "kotti"
```
iamdefinitelyahuman:
```suggestion
        "kotti",
```

PR #355, opened 2020-02-08
bugfix: REVERT outside of function when generating revert map

iamdefinitelyahuman:
### What I did
Fix a compilation bug when a `REVERT` or `INVALID` occurs outside of a function.


iamdefinitelyahuman:
cc @michwill, thanks for finding this

PR #363, opened 2020-02-19
a day I'll never get back...

iamdefinitelyahuman:
### What I did
1. Update `.travis.yml` to use bionic linux instead of xenial.  The latest version of `solc` has dependency issues when using travis' xenial build.
2. Fix issue with `pytest` running tests out of sequence using `-k` flag.
3. Handle newly missing source offsets in solc `0.6.3` and a host of related bugs with `revert` as the final statement in a function.
4. Refactor some scary code.  It's still scary, but it's better than it was before.
5. Bump dependency versions.
6. Update changelog, bump version to `1.6.5`.

This was supposed to be a simple update to fix the failing travis build, and then it all kinda snowballed...  if anyone is watching, I'm sorry for the size of this mess of a pull request.


iamdefinitelyahuman:
Failing from something new in `solc-v0.6.3`

PR #379, opened 2020-04-03
Fix a DivisionByZeroError in case there is no line covered in a contract

palkeo:
"brownie test --coverage" crashed because of that when I included a contract that was left untouched for now.


iamdefinitelyahuman:
You're on a roll!

iamdefinitelyahuman:
This is passing all tests except for the broken links in the docs, which will be fixed by 
f9a47d3.  Merging now.

PR #383, opened 2020-04-03
Always minify sources

iamdefinitelyahuman:
### What I did
Make source minification the default behavior. The increase in load time it causes is minimal and the benefit is significant in my mind.  I've been using it for months now without issue.. I think it's time to make it standard.

### How to verify it
Run tests.


iamdefinitelyahuman:
After working on this a bit more, I now think the opposite approach is the one to take.

PR #448, opened 2020-04-26
Fix importing key file error.

### LINKED ISSUE #440: Importing from a Keystore
I am trying to import from a keystore file to test my contract on a test net, so I run the following command:
```bash
brownie accounts import owner ./UTC--...--5ce9454909639D2D17A3F753ce7d93fa0b9aB12E
```
But I got an error:
```python
Brownie v1.7.2 - Python development framework for Ethereum

  File "brownie/_cli/__main__.py", line 58, in main
    importlib.import_module(f"brownie._cli.{cmd}").main()
  File "brownie/_cli/accounts.py", line 41, in main
    fn(*args["<arguments>"])
  File "brownie/_cli/accounts.py", line 89, in _import
    accounts.load(source_path)
  File "brownie/network/account.py", line 121, in load
    raise FileNotFoundError(f"Cannot find {json_file}")
FileNotFoundError: Cannot find /Users/token/UTC--...--5ce9454909639D2D17A3F753ce7d93fa0b9aB12E.json
```
So it seems like it would add  a `.json` suffix to the file, I created my keystore file form the application `MyCrypto`, it does not provide a suffix, and I remembered that if you tried to generate a keystore file from the `geth`, the file do not has a suffix either, so I think for most of cases, the keystore file does not have a suffix, so is there a need to add the suffix?

Skyge:
### What I did

Fix #440

### How I did it

Modify the function `_import` in the file `/brownie/_cli/accounts.py` and the function `load` in the file `/brownie/network/account.py`

### How to verify it

Run test

### Checklist

### DIFF
--- a/brownie/network/account.py
+++ b/brownie/network/account.py
@@ -120,9 +120,9 @@ class Accounts(metaclass=_Singleton):
             if temp_json_file.exists():
                 json_file = temp_json_file
             else:
-                json_file_name = filename.split("/")[-1]
+                json_file_name = json_file.name
                 json_file = base_accounts_path.joinpath(json_file_name)
-                if not str(json_file).endswith(".json"):
+                if json_file.suffix != ".json":
                     json_file = json_file.with_suffix(".json")
                 if not json_file.exists():
                     raise FileNotFoundError(f"Cannot find {json_file}")
--- a/brownie/network/account.py
+++ b/brownie/network/account.py
@@ -108,17 +108,25 @@ class Accounts(metaclass=_Singleton):
 
         Returns:
             Account instance."""
-        project_path = _get_data_folder().joinpath("accounts")
+        base_accounts_path = _get_data_folder().joinpath("accounts")
         if not filename:
-            return [i.stem for i in project_path.glob("*.json")]
+            return [i.stem for i in base_accounts_path.glob("*.json")]
+
         filename = str(filename)
-        if not filename.endswith(".json"):
-            filename += ".json"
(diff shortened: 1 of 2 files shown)


brownie/network/account.py
```diff
@@ -108,17 +108,25 @@ def load(self, filename: str = None) -> Union[List, "LocalAccount"]:
...
+
         if not json_file.exists():
-            json_file = project_path.joinpath(filename)
-            if not json_file.exists():
-                raise FileNotFoundError(f"Cannot find {json_file}")
+            temp_json_file = json_file.with_suffix(".json")
+            if temp_json_file.exists():
+                json_file = temp_json_file
+            else:
+                json_file_name = filename.split('/')[-1]
```
iamdefinitelyahuman, line 123:
```suggestion
                json_file_name = json_file.name
```

`Path` provides an attribute for this.  Manipulating the string with `/` could cause an issue on Windows filesystems.

brownie/network/account.py
```diff
@@ -108,17 +108,25 @@ def load(self, filename: str = None) -> Union[List, "LocalAccount"]:
...
+                json_file = base_accounts_path.joinpath(json_file_name)
+                if not str(json_file).endswith(".json"):
```
iamdefinitelyahuman, line 125:
```suggestion
                if json_file.suffix != ".json":
```

brownie/network/account.py
Skyge: Ohhh, that is good, I will correct it now.

Issue #484, opened 2020-05-04
Expose nonce for transactions


matnad:
### Overview
Expose the nonce parameter to manually set nonce on deployment and transactions.

Use case: Send multiple transactions in the same block with the `--blockTime [num]` flag for ganache-cli

### Specification
See PR

### Dependencies
Will eventually need `--blockTime` to be exposed as well (separate issue).

iamdefinitelyahuman:
Added in #488

PR #485, opened 2020-05-04
Expose nonce to transactions and deployment

matnad:
### What I did
- changed transfer, deploy and transact functionality to expose the nonce parameter
- minor clarification in error messages
- added test cases for manual nonce

doc update is coming!

Related issue: #484

### How I did it
Added nonce as kwarg or dict element when creating a transaction. E.g.:
`accounts[0].transfer(accounts[1], nonce=5)`
`tester.doNothing({"from": accounts[0], "nonce": 5})`

### How to verify it
run tests 

### Checklist

### DIFF
--- a/brownie/network/account.py
+++ b/brownie/network/account.py
@@ -232,6 +232,7 @@ class _PrivateKeyAccount(PublicKeyAccount):
         amount: int = 0,
         gas_limit: Optional[int] = None,
         gas_price: Optional[int] = None,
+        nonce: Optional[int] = None,
     ) -> Any:
         """Deploys a contract.
 
@@ -244,6 +245,7 @@ class _PrivateKeyAccount(PublicKeyAccount):
             amount: Amount of ether to send with transaction, in wei.
             gas_limit: Gas limit of the transaction.
             gas_price: Gas price of the transaction.
+            nonce: Nonce to use for the transaction.
 
         Returns:
             * Contract instance if the transaction confirms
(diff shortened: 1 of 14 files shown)


brownie/network/account.py
```diff
@@ -260,7 +262,7 @@ def deploy(
                 {
                     "from": self.address,
                     "value": Wei(amount),
-                    "nonce": self.nonce,
+                    "nonce": nonce if nonce is not None else self.nonce,
```
iamdefinitelyahuman, line 265: You can reduce this to `nonce or self.nonce` and make it a bit easier to read

brownie/network/contract.py
```diff
@@ -881,11 +888,16 @@ def _get_tx(owner: Optional[AccountsType], args: Tuple) -> Tuple:
...
+    tx = {"from": owner, "value": 0, "gas": None, "gasPrice": None, "nonce": None}
     if args and isinstance(args[-1], dict):
         tx.update(args[-1])
         args = args[:-1]
-        for key, target in [("amount", "value"), ("gas_limit", "gas"), ("gas_price", "gasPrice")]:
+        for key, target in [
+            ("amount", "value"),
+            ("gas_limit", "gas"),
+            ("gas_price", "gasPrice"),
+            ("nonce", "nonce"),
```
iamdefinitelyahuman, line 899: This logic is translating field names from those used by `brownie` into those expected by `web3py`. Because we don't need to translate the `nonce` key, it shouldn't need to be included here.

Unrelated to your PR.. this function could also be cleaned up a lot.

brownie/network/account.py
matnad: This will fail tests if manual nonce is set to 0, i think enforcing `None` here is correct

brownie/network/contract.py
matnad: Good catch, I missed the update. As long as both the original and translated variables are needed in the dict i think this fct is fine.

matnad:
This is ready for review @iamdefinitelyahuman 
Somehow couldn't resolve that one conflict with rebase o.O

brownie/network/account.py
iamdefinitelyahuman: Ahh good point.

matnad:
closed in favour of #488

Issue #499, opened 2020-05-06
Expand logic for suppressing warnings


iamdefinitelyahuman:
### Overview
Users should not be exposed to warnings that are raised by code outside of their test cases.

Currently, running `brownie test` raises deprecation warnings related to `eth_utils`. The issue was fixed in [`ethereum/eth-utils#186`](https://github.com/ethereum/eth-utils/pull/186) but a new release hasn't been cut yet. Messages such as this could be confusing to new users.

### Specification
1. Expand [`BrownieEnvironmentWarning`](https://github.com/iamdefinitelyahuman/brownie/blob/master/brownie/exceptions.py#L132) into several, more specific warnings.
2. Replace all instances of base warnings with Brownie-specific warnings.
3. Expand the logic in `test/managers/base.py` to suppress all warnings that are not specific to Brownie. This is handled on lines [109](https://github.com/iamdefinitelyahuman/brownie/blob/master/brownie/test/managers/base.py#L109) and [137](https://github.com/iamdefinitelyahuman/brownie/blob/master/brownie/test/managers/base.py#L137).
4. The `--disable-warnings` flag should still suppress Brownie specific warnings. Another flag could be added to stop suppression of other warnings.

carver:
FYI, eth-utils v1.9.0 is out, with that deprecation warning squashed.

PR #501, opened 2020-05-06
Cmd settings to config

matnad:
### What I did

Exposed `block_time`, `default_balance` and `time` ganache-cli parameters
Added cmd_settings to project config file
Added Wei.to() to convert wei to other units

### How I did it
Project settings update network settings on project load
Wei.to() operates with Brownie's Fixed class

### How to verify it
* Add cmd_settings to project config file and lauch cmd
* run tests

### Checklist


iamdefinitelyahuman:
Rebasing in #503 should fix the failing tests.

brownie/network/rpc.py
```diff
@@ -81,8 +87,15 @@ def launch(self, cmd: str, **kwargs: Dict) -> None:
         kwargs.setdefault("evm_version", EVM_DEFAULT)  # type: ignore
         if kwargs["evm_version"] in EVM_EQUIVALENTS:
             kwargs["evm_version"] = EVM_EQUIVALENTS[kwargs["evm_version"]]  # type: ignore
+        kwargs = _validate_cmd_settings(kwargs)
         for key, value in [(k, v) for k, v in kwargs.items() if v]:
-            cmd += f" {CLI_FLAGS[key]} {value}"
+            try:
+                cmd += f" {CLI_FLAGS[key]} {value}"
+            except KeyError:
+                warnings.warn(
```
iamdefinitelyahuman, line 95: In the spirit of #499, this should probably be a brownie-specific warning rather than a builtin.

brownie/network/rpc.py
```diff
@@ -98,6 +111,7 @@ def launch(self, cmd: str, **kwargs: Dict) -> None:
             if web3.isConnected():
                 self._reset_id = self._current_id = self._snap()
                 _notify_registry(0)
+                self._time_offset = self._request("evm_increaseTime", [0])
```
iamdefinitelyahuman, line 114: What is the purpose of this?

brownie/network/rpc.py
```diff
@@ -372,3 +387,38 @@ def _check_connections(proc: psutil.Process, laddr: Tuple) -> bool:
...
+        "evm_version": str,
+        "mnemonic": str,
+        "account_keys_path": str,
+        "fork": str,
+    }
+    for cmd, value in cmd_settings.items():
+        if (
+            cmd in CLI_FLAGS.keys()
+            and cmd in CMD_TYPES.keys()
+            and not type(value) == CMD_TYPES[cmd]
```
iamdefinitelyahuman, line 408: can this be done using `isinstance()` instead of `type()`?

brownie/network/rpc.py
```diff
@@ -372,3 +387,38 @@ def _check_connections(proc: psutil.Process, laddr: Tuple) -> bool:
...
+        ):
+            raise ValueError(
+                f'Wrong type for cmd_settings "{cmd}" ({value}). '
+                f"Found {type(value)}, but expected {CMD_TYPES[cmd]}."
```
iamdefinitelyahuman, line 412: when outputting a type name, the error string is easier to read if you use the  `__name__` member instead of the `repr` for the type itself.
```suggestion
                f"Found {type(value).__name__}, but expected {CMD_TYPES[cmd].__name__}."
```

brownie/network/rpc.py
```diff
@@ -372,3 +387,38 @@ def _check_connections(proc: psutil.Process, laddr: Tuple) -> bool:
...
+            )
+
+    if "default_balance" in cmd_settings:
+        try:
+            cmd_settings["default_balance"] = int(cmd_settings["default_balance"])
```
iamdefinitelyahuman, line 417: Make sure this is well documented, since everywhere else in the config we assume an integer value to be denominated in wei.

tests/conftest.py
```diff
@@ -185,6 +185,17 @@ def testproject(_project_factory, project, tmp_path):
     return project.load(path, "TestProject")
 
 
+@pytest.fixture
```
iamdefinitelyahuman, line 188: given that this fixture is only used in `test_brownie_config`, I think it would be better located in that file than in the `conftest.py`.

tests/data/brownie-test-config.yaml
```diff
@@ -0,0 +1,43 @@
+networks:
```
iamdefinitelyahuman, line 1: Same with this YAML re: location.  I want to try to move away from a `data/` folder and keep source info within the tests where it's relevant, I think it's more readable.  Could you move this to a string at the top of `tests/project/test_brownie_config.py` and add it via `yaml.dump` instead of `shutil.copy`?  This should make the test cases much easier to read since we only have to look at one file instead of three.

tests/project/test_brownie_config.py
```diff
@@ -0,0 +1,60 @@
...
+def test_rpc_project_cmd_settings(devnetwork, testprojectconfig, config, settings_proj):
+    """Test if project specific settings are properly passed on to the RPC."""
+    if devnetwork.rpc.is_active():
+        devnetwork.rpc.kill()
+    cmd_settings_proj = settings_proj["cmd_settings"]
+    testprojectconfig.load_config()
+    devnetwork.connect("development")
+
+    # Check if rpc time is roughly the start time in the config file
+    # Use diff < 25h to dodge potential timezone differences
```
iamdefinitelyahuman, line 45: nice :)

brownie/convert/datatypes.py
```diff
@@ -74,6 +74,12 @@ def __add__(self, other: Any) -> "Wei":
     def __sub__(self, other: Any) -> "Wei":
         return Wei(super().__sub__(_to_wei(other)))
 
+    def to(self, unit: str) -> "Fixed":
+        try:
```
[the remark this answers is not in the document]
matnad: absolutely

brownie/convert/datatypes.py
```diff
@@ -74,6 +74,12 @@ def __add__(self, other: Any) -> "Wei":
...
+            return Fixed(self * Fixed(10) ** -UNITS[unit])
+        except KeyError:
+            raise TypeError(f'Cannot convert wei to unknown unit: "{unit}". ')
```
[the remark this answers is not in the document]
matnad: Today I learned... :)

brownie/network/rpc.py
matnad: As far as I could see, "evm_increaseTime" is the only way to get the time offset out of ganache. This is essentially `sleep 0 and return the offset` to make sure the `Rpc.time()` is in sync with ganache from the start. If we omit it, then `Rpc.time()` is equal to the current time until we use `Rpc.sleep()` the first time.

brownie/network/rpc.py
matnad: yes :)

brownie/network/rpc.py
matnad: Ah I was looking for something like this, thx!

brownie/network/rpc.py
matnad: The reasoning behind this is that ganache wants the argument in ether. I wanted to keep the parameters as close to the ganache-cli parameters as possible. It's your call if you want to keep a strict "for brownie, input in ether is in wei units" approach. Let me know

tests/data/brownie-test-config.yaml
matnad: Jep, sounds good!
Will change this and move the fixture.

brownie/network/rpc.py
iamdefinitelyahuman: I agree with your approach, I just wanted to stress the importance of documentation on this :)

brownie/data/default-config.yaml
```diff
@@ -12,6 +12,7 @@ networks:
         gas_price: 0
         reverting_tx_gas_limit: 6721975
         default_contract_owner: true
+        cmd_settings: {}
```
iamdefinitelyahuman, line 15: Given this field is empty by default and you are checking for it's existence prior to working with it, is there a need to include it in the default config?

tests/project/test_brownie_config.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    with testproject._path.joinpath("brownie-config.yaml").open("w") as fp:
+        yaml.dump(yaml.load(test_brownie_config), fp)
+
+    # Load the networks.development config from the created file and yield it
+    with testproject._path.joinpath("brownie-config.yaml").open() as fp:
+        conf = yaml.safe_load(fp)["networks"]["development"]
+        yield conf
+
+    os.remove(testproject._path.joinpath("brownie-config.yaml"))
```
iamdefinitelyahuman, line 41: `testproject` creates a fresh project in a new folder for every test that uses it, so there's no need for cleaning up the filesystem like this.

tests/project/test_brownie_config.py
```diff
@@ -0,0 +1,115 @@
...
+                evm_version: byzantium
+                mnemonic: brownie2
+    """
+    with testproject._path.joinpath("brownie-config.yaml").open("w") as fp:
+        yaml.dump(yaml.load(test_brownie_config), fp)
+
+    # Load the networks.development config from the created file and yield it
+    with testproject._path.joinpath("brownie-config.yaml").open() as fp:
+        conf = yaml.safe_load(fp)["networks"]["development"]
+        yield conf
```
iamdefinitelyahuman, line 39: does this yield need to be inside the `open` context manager? I feel like that could have some weird unintended side effects.

tests/convert/test_wei.py
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/python3
+import pytest
 
+from brownie import Fixed
```
iamdefinitelyahuman, line 4: probably better to import both `Fixed` and `Wei` from where they are actually declared, `brownie.convert.datatypes`

brownie/data/default-config.yaml
matnad: The title in the docs says the default config shows all the available settings with their default values. I was on the fence about including it, but then the description in the docs would have to be changed slightly. Your call!

tests/project/test_brownie_config.py
matnad: I got this from the pytest docs: https://docs.pytest.org/en/2.8.7/yieldfixture.html
I realize now this is from quite an outdated version of pytest and the behaviour is changed slightly. I still assume the file will be closed at the end of the test like this. I can change it tho to be sure :)

brownie/data/default-config.yaml
iamdefinitelyahuman: Ahh very good point. Ok then..  can we make the setting `null` rather than `{}`? I think i just don't like the squiggly brackets.

PR #528, opened 2020-05-18
contract_strategy

iamdefinitelyahuman:
### What I did
Add `contract_strategy` - a hypothesis strategy to draw from `ContractContainer` objects.

### How I did it
Similar to `address`, it uses a `DeferredStrategy` and checks for contracts in active projects when it's accessed.

### How to verify it
Run tests.

### DIFF
--- a/tests/test/strategies/test_contract_strategy.py
+++ b/tests/test/strategies/test_contract_strategy.py
@@ -1,0 +1,26 @@
+#!/usr/bin/python3
+
+import pytest
+from hypothesis import given
+from hypothesis.strategies._internal.deferred import DeferredStrategy
+
+from brownie.network.contract import ProjectContract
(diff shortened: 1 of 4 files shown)


tests/test/strategies/test_contract_strategy.py
```diff
@@ -0,0 +1,26 @@
...
+def test_repr():
+    assert repr(contract_strategy("Token")) == "sampled_from(Token)"
+
+
+def test_does_not_exist():
+    with pytest.raises(NameError):
+        contract_strategy("Foo").validate()
+
+
+@given(contract=contract_strategy("BrownieTester"))
```
fubuloubu, line 24: probably could add number of samples = 1 to a settings object here

PR #530, opened 2020-05-20
Use `web3.eth.generateGasPrice` for gas price

iamdefinitelyahuman:
### What I did
Allow use of custom gas strategies via the web3 gas price API

### How I did it
* When determining gas price, use `web3.eth.generateGasPrice` if no value is given in the config.
* Set the default strategy to check `eth_gasPrice`

### How to verify it
Run tests.


michwill:
Is it possible to choose between different gas strategies? E.g. fast/medium/slow/glacier in web3py

PR #551, opened 2020-05-24
fix: check web3.eth.accounts when calling Accounts.at

iamdefinitelyahuman:
### What I did
Check `web3.eth.accounts` when accessing an account via `Accounts.at` - useful if a new account was added from e.g. a hardware wallet being unlocked.

### How to verify it
Run tests.

### DIFF
--- a/brownie/network/account.py
+++ b/brownie/network/account.py
@@ -152,10 +152,15 @@ class Accounts(metaclass=_Singleton):
             Account instance.
         """
         address = _resolve_address(address)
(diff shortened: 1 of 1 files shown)


BrokedTV:
When i add an account with 
accounts.at('0x5ba30f12022c0b29acee4faed0666dcf6524f424', force=True)
on a development network, I can see it added to "accounts"
but if i call web3.eth.accounts it's not added.
how could i sync them?

PR #564, opened 2020-05-28
Gas trace

### LINKED ISSUE #563: Improve tx.call_trace with gas report
### Overview

Show gas used for each function and call when using `tx.call_trace()`

### Specification
see PR

matnad:
### What I did
Add gas info to tx.call_trace()

Closes #563 

### How I did it
Build gas info from the trace object

### How to verify it
`tx.call_trace()`

### Checklist

### DIFF
--- a/brownie/network/transaction.py
+++ b/brownie/network/transaction.py
@@ -616,8 +616,8 @@ class TransactionReceipt:
         print(result)
 
     def _get_trace_gas(self, start: int, stop: int) -> Tuple[int, int]:
-        totalGas = 0
-        internalGas = 0
+        total_gas = 0
+        internal_gas = 0
         is_internal = True
         trace = self.trace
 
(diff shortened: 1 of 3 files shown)


brownie/network/transaction.py
```diff
@@ -615,6 +615,49 @@ def info(self) -> None:
                     result += f"\n      {key}: {color('bright blue')}{value}{color}"
         print(result)
 
+    def _get_trace_gas(self, start: int, stop: int) -> Tuple[int, int]:
+        total_gas = 0
+        internal_gas = 0
+        is_internal = True
+        trace = self.trace
+
+        is_gas_forwarded = trace[start]["depth"] > trace[start - 1]["depth"]
```
iamdefinitelyahuman, line 624: What if `start == 0` ?

matnad: You are right, `start = 0` will compare the first to the last element. Since the last and first step always have the same depth and "from 0" is never an external call, this would work, but I added a check for `start > 0` to be more precise and a also inlined this variable, since it ended up being used only once.

PR #577, opened 2020-05-30
Dev artifacts

### LINKED ISSUE #576: Add deployment artifacts in development networks
### Overview
Save deployment artifacts for development networks so they can be imported by other programs
Add a deployment map to keep track of deployed artifacts

### Specification
Save deployment artifacts for development networks at `build/deployments/dev/`
Maintain a deployment map (of all deployments) at `build/depoyments/map.json` in the form of
```
{
  "chainid": { 
    "contractName": [
      "contract address 1", 
      "contract address 2"
    ]
  }
```
Where the addresses for each contract are sorted with the most recent deployment first.

### Dependencies
A custom build folder is also needed for this to fully support other programs.

matnad:
### What I did

* Add a setting to save deployment artifacts in development networks
* Add a deployment map that keeps track of deployed contracts, sorted by time

closes #576

### How I did it
Save deployment artifacts for development networks at build/deployments/dev/
Maintain a deployment map (of all deployments) at build/depoyments/map.json in the form of
```
{
  "chainid": { 
    "contractName": [
      "contract address 1", 
      "contract address 2"
    ]
  }
```

### How to verify it
Deploy a contract and check `build/deployments`

### Checklist

### DIFF
--- a/brownie/network/rpc.py
+++ b/brownie/network/rpc.py
@@ -14,6 +14,7 @@ from urllib.parse import urlparse
 
 import psutil
 
+import brownie
 from brownie._config import EVM_EQUIVALENTS
 from brownie._singleton import _Singleton
 from brownie.convert import Wei
@@ -353,6 +354,8 @@ class Rpc(metaclass=_Singleton):
         self._undo_buffer.clear()
         self._redo_buffer.clear()
         self._snapshot_id = self._current_id = self._revert(self._snapshot_id)
+        for project in brownie.project.get_loaded_projects():
+            project._clear_dev_deployments(web3.eth.blockNumber)
         return f"Block height reverted to {web3.eth.blockNumber}"
 
     def reset(self) -> str:
@@ -365,6 +368,8 @@ class Rpc(metaclass=_Singleton):
         self._undo_buffer.clear()
         self._redo_buffer.clear()
         self._reset_id = self._current_id = self._revert(self._reset_id)
+        for project in brownie.project.get_loaded_projects():
+            project._clear_dev_deployments()
         return f"Block height reset to {web3.eth.blockNumber}"
 
 
--- a/brownie/network/rpc.py
+++ b/brownie/network/rpc.py
@@ -14,7 +14,6 @@ from urllib.parse import urlparse
 
 import psutil
 
-import brownie
 from brownie._config import EVM_EQUIVALENTS
 from brownie._singleton import _Singleton
 from brownie.convert import Wei
@@ -354,8 +353,6 @@ class Rpc(metaclass=_Singleton):
         self._undo_buffer.clear()
         self._redo_buffer.clear()
         self._snapshot_id = self._current_id = self._revert(self._snapshot_id)
-        for project in brownie.project.get_loaded_projects():
-            project._clear_dev_deployments(web3.eth.blockNumber)
         return f"Block height reverted to {web3.eth.blockNumber}"
 
     def reset(self) -> str:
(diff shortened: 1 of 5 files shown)


brownie/project/main.py
```diff
@@ -207,6 +207,7 @@ def load(self) -> None:
         self._compile(changed, self._compiler_config, False)
         self._compile_interfaces(interface_hashes)
         self._create_containers()
+        self._clear_dev_deployments()
```
matnad, line 210: Not sure this is the right location to do this. 
Any suggestions?

brownie/project/main.py
```diff
@@ -292,30 +292,116 @@ def _compile_interfaces(self, compiled_hashes: Dict) -> None:
...
             container._contracts.append(contract)
 
+            contract_address = build_json.stem
+            if chainid in deployment_map:
+                if contract_name in deployment_map[chainid] and isinstance(
+                    deployment_map[chainid][contract_name], list
+                ):
+                    deployment_map[chainid][contract_name].insert(0, contract_address)
+                else:
+                    deployment_map[chainid][contract_name] = [contract_address]
```
iamdefinitelyahuman, line 329: This could all be handled as one line..  the way it is now I find a bit hard to grok.

```python
deployment_map[chainid].setdefault(contract_name, []).insert(0, contract_address)
```

Also, am I correct that inserting at the start means this is sequenced newest to oldest? What is the reasoning there?

brownie/project/main.py
```diff
@@ -292,30 +292,116 @@ def _compile_interfaces(self, compiled_hashes: Dict) -> None:
...
+                deployment_map[chainid] = {contract_name: [contract_address]}
+        with self._path.joinpath("build/deployments/map.json").open("w") as fp:
+            json.dump(deployment_map, fp, sort_keys=True, indent=2, default=sorted)
+
+    def _clear_dev_deployments(self, after_block: int = 0) -> None:
+        path = self._path.joinpath(f"build/deployments/dev")
+        if path.exists():
+            deployments = list(path.glob("*.json"))
+            deployment_map = self._load_deployment_map()
+            for deployment in deployments:
```
iamdefinitelyahuman, line 340: `deployment` and `deployments` are very similar names.  Is there a need to create `deployments`? You could iterate directly from `path.glob('*.json')`

brownie/project/main.py
```diff
@@ -292,30 +292,116 @@ def _compile_interfaces(self, compiled_hashes: Dict) -> None:
...
+                if after_block == 0:
+                    deployment.unlink()
+                    if "dev" in deployment_map:
+                        del deployment_map["dev"]
+                    self._save_deployment_map(deployment_map)
```
iamdefinitelyahuman, line 345: This could happen just once, after the iteration is finished

brownie/network/rpc.py
```diff
@@ -352,6 +353,8 @@ def revert(self) -> str:
         self._undo_buffer.clear()
         self._redo_buffer.clear()
         self._snapshot_id = self._current_id = self._revert(self._snapshot_id)
+        for project in brownie.project.get_loaded_projects():
```
iamdefinitelyahuman, line 356: This feels out of place to me.  In general, I think `project` should be importing and interacting with `network`, but not the other way around.

Maybe instead add `_revert` and `_reset` methods to `Project`, and call `rpc._register_revert` during `Project.load`?  And then handle the `_clear_dev_deployments` logic in those functions.

brownie/project/main.py
matnad: This is brilliant! I have not heard of setdefault befiore.

It being ordered from newest to oldest is simply for convenience reasons that `contract[0]` is always the newest contract instead of `contract[contract.length - 1]` (yay javascript)

I think the performance impact to shift the list is neglible (and using deque is probably overkill). However, if you think ordering from last to first is more intuitivie I am open to change it

iamdefinitelyahuman: no i'm totally fine with it, just didn't see the logic.  I forgot JS doesn't allow negative indexing..  that's a very good reason to prefer to access from the left :)

matnad:
closed in favor of new PR

PR #581, opened 2020-05-31
Configurable project structure

### LINKED ISSUE #578: Customize project directory structure
### Overview
It should be possible to use a different folder structure within a Brownie project.

### Specification
Add the following default config settings:

```python
project_structure:
    build: "./build"
    contracts: "./contracts"
    interfaces: "./interfaces"
    reports: "./reports"
    scripts: "./scripts"
    tests: "./tests"
```

When loading a project, search for a config file in that folder. If one is found and any of these fields are present, use them rather than the defaults.

iamdefinitelyahuman:
### What I did
Add config settings to modify the directory structure of a project.

Closes #578

### How I did it
* Add a config section `project_structure` which defines the expected directories.
* In `_config`, add a function `_load_project_structure_config` that checks a folder for a config file, and if yes returns the custom structure. If none is found it returns the default values.  This is necessary because we check for the existence of a brownie project based on the directory structure - so the settings must be loaded _prior_ to actually loading the project, in order to see that there is a project to load.
* Modified references to absolute directory names with references to the config settings throughout the code base.

### How to verify it
Run the tests. I added a couple of cases to confirm the basic functionality and have been using it locally to see that it works.  This is difficult to test extensively without simply parametrizing the whole test suite :unamused: 

### Checklist

### DIFF
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -47,17 +47,6 @@ from brownie.project.ethpm import get_deployment_addresses, get_manifest
 from brownie.project.sources import Sources, get_pragma_spec
 
 BUILD_FOLDERS = ["contracts", "deployments", "interfaces"]
-FOLDERS = [
-    "contracts",
-    "interfaces",
-    "scripts",
-    "tests",
-    "reports",
-    "build",
-    "build/contracts",
-    "build/deployments",
-    "build/interfaces",
-]
 MIXES_URL = "https://github.com/brownie-mix/{}-mix/archive/master.zip"
 
 GITIGNORE = """__pycache__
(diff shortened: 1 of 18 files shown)


brownie/project/main.py
```diff
@@ -409,9 +417,21 @@ def check_for_project(path: Union[Path, str] = ".") -> Optional[Path]:
...
+            (
+                i
+                for i in folder.joinpath(contracts_path).glob("**/*")
+                if i.suffix in (".vy", ".sol")
+            ),
+            None,
+        ):
             return folder
-        if folder.joinpath("contracts").is_dir() and folder.joinpath("tests").is_dir():
+        if folder.joinpath(tests_path).is_dir() and folder.joinpath(tests_path).is_dir():
```
matnad, line 434: The first check should be `contracts_path`

brownie/project/main.py
```diff
@@ -37,6 +38,7 @@
 from brownie.project.ethpm import get_deployment_addresses, get_manifest
 from brownie.project.sources import Sources, get_pragma_spec
 
+BUILD_FOLDERS = ["contracts", "deployments", "interfaces"]
 FOLDERS = [
```
matnad, line 42: FOLDERS is no longer used i think

iamdefinitelyahuman: good catch!

brownie/project/main.py
iamdefinitelyahuman: Fixed it, and refactored a bit to make that `next` statement less ugly.  I hadn't seen what `black` did to it :eyes:

PR #589, opened 2020-06-01
Ci external doctest

iamdefinitelyahuman:
### What I did
Move the check for external doc links to it's own job and mark that job as experimental.

### How to verify it
If this PR shows as passing all checks, it worked.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -13,8 +13,13 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman:
I can't find a way to allow `docs-external` to show as failing, without also failing the entire build.  So right now it's running, but always shows as passed even if it finds an issue.

Will have to remember to look periodically and see if anything is going wrong.  Given we can't control the uptime on external links though, I think this is preferrable to an unfixable, failing build.

PR #593, opened 2020-06-02
Feat tx required confs

### LINKED ISSUE #587: User-defined wait time for broadcasted transactions
### Overview
Allow users to specify the number of confirmations they wish to wait for a transaction.

### Specification
When broadcasting a transaction, the `required_confs` kwarg can determine how many confirmations to wait for.  Value can be 0 for a non-blocking action that returns a pending `TransactionReceipt`.  The default value is 1.

```python
accounts[0].transfer(accounts[1], required_confs=n)
Token.transfer(accounts[1], 10000, {'from': accounts[0], 'required_confs': n)
```

Once a tx is broadcast, it should also be possible to wait for confirmations via a `wait` method.  If the number of confirmations is already greater than `n`, nothing happens.

```python
tx.wait(n)
```

Once this is added, the old and half-broken logic around `CTRL-C` in the console can also be removed.

matnad:
### What I did

Add `required_confs` as an argument to transactions with a default value of 1 (same as current behaviour).

Setting `required_confs = n > 1` will wait for n blocks to be mined before the transaction receipt is processed.

If during the confirmation time the transaction is dropped (uncle block), the transaction will wait until it is picked up again and has the required amount of confirmations.

if `required_confs == 0` the transaction will immediately return a pending tx receipt and not wait for a confirmation.

Add `tx.wait(n)` to retroactively add required_confs and wait until they are confirmed.

TODO: Still needs more documentation

Closes #587

### How I did it
Added the argument to all deply, transfer and transact functions.

Refactored the confirmation logic

### How to verify it
I added tests for this.
or try `accounts[0].transfer(accounts[1], "1 ether", {'required_confs': 3})`

### Checklist

### DIFF
--- a/brownie/network/transaction.py
+++ b/brownie/network/transaction.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-
+import sys
 import threading
 import time
 from hashlib import sha1
@@ -114,6 +114,7 @@ class TransactionReceipt:
         "logs",
         "nonce",
         "receiver",
+        "required_confs",
         "sender",
         "status",
         "txid",
@@ -163,6 +164,7 @@ class TransactionReceipt:
         self.status = -1
         self.txid = txid
         self.fn_name = name
+        self.required_confs = 1
 
         if name and "." in name:
             self.contract_name, self.fn_name = name.split(".", maxsplit=1)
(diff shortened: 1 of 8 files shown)


brownie/network/transaction.py
```diff
@@ -163,6 +171,7 @@ def __init__(
         self.status = -1
         self.txid = txid
         self.fn_name = name
+        self.required_confs = required_confs
```
iamdefinitelyahuman, line 174: Do you think this is important to be part of the public API? Given it only determines how long the initial call is blocking for, I can't really see why a user would want to reference it later.

tests/network/transaction/test_required_confirmations.py
```diff
@@ -0,0 +1,83 @@
+import time
+
+import pytest
+
+# @pytest.fixture(scope="module")
```
iamdefinitelyahuman, line 5: This can go.

brownie/network/transaction.py
matnad: `tx.wait()` will set this attribute, so I thought it might be useful if you can call it before calling `tx.wait()`. But I can see moving this to an internal attribute

PR #595, opened 2020-06-03
Project installation fixes

### LINKED ISSUE #497: Install packages with non-standard folder structures
### Overview

I am wanting to install [kollateral](https://github.com/kollateral/kollateral) as a dependency, but it has a different project structure than brownie expects.

The repo also does not have any git tags.

### Specification

`.../main.py` has `if not install_path.joinpath("contracts").exists():`. For now I've just commented this out, since contracts is just deeper in the project, but I think having this be a configurable value would work well.

For the lack of tags, it would be nice if we could just specify a full git hash.

iamdefinitelyahuman:
### What I did
* Allow installing packages with a non-standard contracts folder - closes #497 
* Allow installing packages that won't compile

### How I did it
* If the project has no `contracts` folder, but there is another folder that contains contract files, create a configuration file with a `project_structure` field to point at the other folder
* If a project cannot compile, raise a warning but do not prevent it from installing. Users can still import individual contracts even if Brownie cannot load the project as a whole.

### How to verify it
Run tests.

### DIFF
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -764,28 +764,29 @@ def _install_from_github(package_id: str) -> str:
     shutil.move(installed, install_path)
 
     try:
-        brownie_config: Dict = {"project_structure": {}}
+        if not install_path.joinpath("brownie-config.yaml").exists():
+            brownie_config: Dict = {"project_structure": {}}
 
-        contract_paths = set(
-            i.relative_to(install_path).parts[0] for i in install_path.glob("**/*.sol")
-        )
-        contract_paths.update(
-            i.relative_to(install_path).parts[0] for i in install_path.glob("**/*.vy")
-        )
-        if not contract_paths:
-            raise InvalidPackage(f"{package_id} does not contain any .sol or .vy files")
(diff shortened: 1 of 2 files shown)


brownie/project/main.py
```diff
@@ -761,14 +764,40 @@ def _install_from_github(package_id: str) -> str:
...
+        elif len(contract_paths) == 1:
+            brownie_config["project_structure"]["contracts"] = contract_paths.pop()
+        else:
+            raise InvalidPackage(
+                f"{package_id} has no `contracts/` subdirectory, and "
+                "multiple directories containing source files"
+            )
+
+        with install_path.joinpath("brownie-config.yaml").open("w") as fp:
+            yaml.dump(brownie_config, fp)
```
matnad, line 788: This will overwrite the config file. Are you sure no other method during installation is interacting with the config file now or in the future?

iamdefinitelyahuman: Oh hrmm..  this is a good point. I was assuming it doesn't exist because I was staring at npm packages for so long. If it _does_ exist, all of this logic is moot.

PR #608, opened 2020-06-05
Fix repeated alerts by using type instead of instanceof

tbrent:
### What I did

Fix repeated alerts

### How I did it

Used `type` instead of `instanceof`, because `bool` is a subclass of `int` in python. 

### How to verify it

It's a small change. Also automated tests should pickup if this is broken. 

### Checklist

- [✅] I have added an entry to the changelog

### DIFF
--- a/brownie/network/alert.py
+++ b/brownie/network/alert.py
@@ -90,7 +90,7 @@ class Alert:
                 start_value = value
                 if not repeat:
                     repeat = None
-                elif type(repeat) == int:
+                elif isinstance(repeat, int) and not isinstance(repeat, bool):
                     repeat -= 1
         finally:
             _instances.discard(self)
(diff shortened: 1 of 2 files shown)


brownie/network/alert.py
```diff
@@ -90,7 +90,7 @@ def _loop(
                 start_value = value
                 if not repeat:
                     repeat = None
-                elif isinstance(repeat, int):
+                elif type(repeat) == int:
```
iamdefinitelyahuman, line 93: It's more verbose, but I think preferable to use `isinstance(repeat, int) and not isinstance(repeat, bool)` - because [`Wei`](https://eth-brownie.readthedocs.io/en/stable/core-types.html#wei) is an `int` subclass and commonly used for return values, I can imagine a situation where the user is trying to set an alert with that type.

tbrent: Makes sense. Updated to reflect this.

I'll keep a lookout for possible improvements while we apply our usecase to it 👍

Issue #568, opened 2020-05-28
Coverage.py plugin for generating coverage reports


iamdefinitelyahuman:
### Overview
Create a [`coverage.py` plugin](https://coverage.readthedocs.io/en/coverage-5.1/api_plugin.html#module-coverage.plugin) to generate coverage reports from Brownie project tests.

### Specification
The plugin must work with all existing Brownie pytest functionality:

* Must work for both Solidity and Vyper source files
* Must be compatible with `xdist`
* Must be handle both statement and branch coverage (although statement can be represented as line if that's easier)
* Must be able to include previously calculated coverage data, when the `--update` flag is used

The plugin should activate using the `--coverage` cli flag when calling `brownie test`. The syntax of the flag can be expanded to allow users to choose which format(s) to output, e.g. XML, HTML, etc.

### Resources
* Coverage data is generated on a per-tx basis and aggregated by functions in [`brownie/test/overage.py`](https://github.com/iamdefinitelyahuman/brownie/blob/master/brownie/test/coverage.py).
* The final report is saved by a pytest hook in one of the [`brownie/test/managers`](https://github.com/iamdefinitelyahuman/brownie/tree/master/brownie/test/managers) modules.
* The data format is specified in the [documentation](https://eth-brownie.readthedocs.io/en/stable/build-folder.html#test-results-and-coverage-data)

### Completion Goals
1. A fully working plugin capable of generating reports in any format produced by `coverage.py`.
2. Test cases that generate and verify XML and HTML outputs using the pytest's `testdir` fixture. See the existing [plugin tests](https://github.com/iamdefinitelyahuman/brownie/tree/master/tests/test).

Ignas:
https://raw.githubusercontent.com/jenkinsci/cobertura-plugin/master/src/test/resources/hudson/plugins/cobertura/coverage-with-data.xml - Cobertura file example

kovalevvlad:
Another use-case for this - show test coverage on Gitlab PRs

PR #632, opened 2020-06-17
PoC of coverage.py integration

Ignas:
### What I did

Implemented a proof of concept for using python coverage module to do coverage tracking and reporting

Related issue: #568

### How I did it

So the way coverage.py does coverage does not really match the existing architecture, or even the idea that single function call can produce a whole set of covered lines. Thus I had to "improvise". The magic is in the changes made to `transaction.py`, basically `coverage.py` expects to have a **frame** for every **segment** of **lines** covered, so we just wait until transaction is over (I am not familiar with the code or etherium, so I had to make do) and run a function for every trace in the transaction, this way our plugin can hook into that function, and pretend it was interpreting the contract all along.

### How to verify it

So I mounted https://github.com/matnad/liquid-gas-token into the docker compose thing, and ran:

```
brownie test -C tests/unit/liquidity/test_buy_lgt.py
```

### Checklist

Nothing done, because this is just a PoC

### Things missing

- So the way coverage does branch coverage is quite different from the current approach. Coverage.py collects all the lines covered, then it does analysis during which it figures out all the jumps, so what is needed is an `arcs()` function that would look at coverageMap and list all the arcs.

- Currently we collect coverage data using offsets, and then convert them to lines, this might cause problems, but I am not sure

- The coverage unit is a pair of "contractName" and "functionName" and this is where my lack of understanding about how the machine works really shines through - we have to figure out all the lines being covered during execution, and I really wanted to avoid going back to the _build to check if functions are in the files, which leads to me being unable to figure out how to reject functions that do not exist but are in the trace. For example:

```
  contract: LiquidGasToken - 25.1%
    ERC20PointerSupply.balanceOf - 100.0%
    LiquidERC20.<receive> - 100.0%
    LiquidERC20.ethToTokenInput - 100.0%
    LiquidERC20.ethToTokenOutput - 100.0%
    LiquidERC20.ethToTokenSwapInput - 100.0%
    LiquidERC20.ethToTokenSwapOutput - 100.0%
    LiquidERC20.ethToTokenTransferInput - 100.0%
    LiquidERC20.ethToTokenTransferOutput - 100.0%
    LiquidERC20.getEthToTokenInputPrice - 100.0%
    LiquidERC20.getEthToTokenOutputPrice - 100.0%
    LiquidERC20.getInputPrice - 100.0%
    LiquidERC20.getOutputPrice - 100.0%
    LiquidGasToken.mint - 100.0%
    LiquidERC20.addLiquidity - 75.0%
    ERC20PointerSupply._approve - 0.0%
    ERC20PointerSupply._transfer - 0.0%
    ERC20PointerSupply.allowance - 0.0%
    ERC20PointerSupply.approve - 0.0%
    ERC20PointerSupply.decreaseAllowance - 0.0%
    ERC20PointerSupply.increaseAllowance - 0.0%
    ERC20PointerSupply.ownedSupply - 0.0%
    ERC20PointerSupply.totalSupply - 0.0%
    ERC20PointerSupply.transfer - 0.0%
    ERC20PointerSupply.transferFrom - 0.0%
    LiquidERC20._transferToSelf - 0.0%
    LiquidERC20.getTokenToEthInputPrice - 0.0%
    LiquidERC20.getTokenToEthOutputPrice - 0.0%
    LiquidERC20.poolBalanceOf - 0.0%
    LiquidERC20.poolTokenReserves - 0.0%
    LiquidERC20.poolTotalSupply - 0.0%
    LiquidERC20.poolTransfer - 0.0%
    LiquidERC20.removeLiquidity - 0.0%
    LiquidERC20.tokenToEthInput - 0.0%
    LiquidERC20.tokenToEthOutput - 0.0%
    LiquidERC20.tokenToEthSwapInput - 0.0%
    LiquidERC20.tokenToEthSwapOutput - 0.0%
    LiquidERC20.tokenToEthTransferInput - 0.0%
    LiquidERC20.tokenToEthTransferOutput - 0.0%
    LiquidGasToken.buyAndFree - 0.0%
    LiquidGasToken.buyAndFree22457070633 - 0.0%
    LiquidGasToken.buyMaxAndFree - 0.0%
    LiquidGasToken.create2 - 0.0%
    LiquidGasToken.deploy - 0.0%
    LiquidGasToken.free - 0.0%
    LiquidGasToken.freeFrom - 0.0%
    LiquidGasToken.mintFor - 0.0%
    LiquidGasToken.mintToLiquidity - 0.0%
    LiquidGasToken.mintToSell - 0.0%
    LiquidGasToken.mintToSell9630191 - 0.0%
    LiquidGasToken.mintToSellTo - 0.0%
    LiquidGasToken.mintToSellTo25630722 - 0.0%

Coverage report saved at /liquid-gas-token/reports/coverage.json
View the report using the Brownie GUI

---------- coverage: platform linux, python 3.6.10-final-0 -----------
Name                                                               Stmts   Miss  Cover
--------------------------------------------------------------------------------------
contracts/LiquidGasToken/ERC20PointerSupply.balanceOf                  1      0   100%
contracts/LiquidGasToken/LiquidERC20.addLiquidity                     12      0   100%
contracts/LiquidGasToken/LiquidERC20.ethToTokenInput                   6      0   100%
contracts/LiquidGasToken/LiquidERC20.ethToTokenOutput                  7      0   100%
contracts/LiquidGasToken/LiquidERC20.ethToTokenSwapInput               1      0   100%
contracts/LiquidGasToken/LiquidERC20.ethToTokenSwapOutput              1      0   100%
contracts/LiquidGasToken/LiquidERC20.ethToTokenTransferInput           3      0   100%
contracts/LiquidGasToken/LiquidERC20.ethToTokenTransferOutput          3      0   100%
contracts/LiquidGasToken/LiquidERC20.getEthToTokenInputPrice           1      0   100%
contracts/LiquidGasToken/LiquidERC20.getEthToTokenOutputPrice          1      0   100%
contracts/LiquidGasToken/LiquidERC20.getInputPrice                     1      0   100%
contracts/LiquidGasToken/LiquidERC20.getOutputPrice                    1      0   100%
contracts/LiquidGasToken/LiquidGasToken                                0      0   100%
contracts/LiquidGasToken/LiquidGasToken._createContracts               0      0   100%
contracts/LiquidGasToken/LiquidGasToken.addLiquidity                   0      0   100%
contracts/LiquidGasToken/LiquidGasToken.balanceOf                      0      0   100%
contracts/LiquidGasToken/LiquidGasToken.ethToTokenSwapInput            0      0   100%
contracts/LiquidGasToken/LiquidGasToken.ethToTokenSwapOutput           0      0   100%
contracts/LiquidGasToken/LiquidGasToken.ethToTokenTransferInput        0      0   100%
contracts/LiquidGasToken/LiquidGasToken.ethToTokenTransferOutput       0      0   100%
contracts/LiquidGasToken/LiquidGasToken.getEthToTokenInputPrice        0      0   100%
contracts/LiquidGasToken/LiquidGasToken.getEthToTokenOutputPrice       0      0   100%
contracts/LiquidGasToken/LiquidGasToken.mint                           3      0   100%
contracts/LiquidGasToken/SafeMath.add                                  1      1     0%
contracts/LiquidGasToken/SafeMath.mul                                  2      2     0%
--------------------------------------------------------------------------------------
TOTAL                                                                 44      3    93%
```

If I am loooking at the trace, I see calls to SafeMath.add and SafeMath.mul attributed to the LiquidGasToken (even the file is LiquidGasToken.sol) but ignoring all the functions not from the same file would lead to `LiquidERC20.addLiquidity` being ignored. Also `addLiquidity` gets a trace for both LiquidERC20 and LiquidGasToken, no idea why or how.

- Reporting in xml/html is not done at all, to do it you would have to convert the line numbers to relative and add `source_token_lines()` and `source()` methods that return the code for a single function, unless you want coverage per file, in which case - the code that interprets the traces has to get updated to do files.

### DIFF
--- a/brownie/exceptions.py
+++ b/brownie/exceptions.py
@@ -3,8 +3,10 @@
 import sys
 from typing import Optional, Type
 
+import eth_abi
 import psutil
 import yaml
+from hexbytes import HexBytes
 
 import brownie
 
@@ -82,7 +84,20 @@ class VirtualMachineError(Exception):
             if "data" not in exc:
                 raise ValueError(exc["message"]) from None
 
-            self.message: str = exc["message"]
+            self.message: str = exc["message"].rstrip(".")
+
+            if isinstance(exc["data"], str):
+                # handle parity exceptions - this logic probably is not perfect
+                if "0x08c379a0" in exc["data"]:
+                    revert_type, err_msg = [i.strip() for i in exc["data"].split("0x08c379a0", 1)]
+                    err_msg = eth_abi.decode_abi(["string"], HexBytes(err_msg))
+                    err_msg = f"{revert_type} '{err_msg}'"
+                elif exc["data"].endswith("0x"):
+                    err_msg = exc["data"][:-2].strip()
+                else:
+                    err_msg = exc["data"]
+                raise ValueError(f"{self.message}: {err_msg}") from None
+
             try:
                 txid, data = next((k, v) for k, v in exc["data"].items() if k.startswith("0x"))
             except StopIteration:


iamdefinitelyahuman:
Hey, awesome!  Thanks for working on this :pray: 

The overlap of functions that you're seeing is a result of [inheritance](https://solidity.readthedocs.io/en/v0.6.10/contracts.html#inheritance).  Multiple inheritance is fairly common in Solidity, so to give a more detailed report Brownie is referencing functions according to the final, compiled contract that they're found in.  `contractName` is the name of the contract that was compiled, `functionName` is formatted as `[contract].[function]` where `[contract]` is the name of the actual contract that contains the source code for `[function]`.


I hadn't considered the challenge this creates.  `coverage.py` wants to handle things on a per-file basis, but if we merge all the coverage info based on the actual file we're not giving as complete of a picture.

Perhaps we could mutate how the results are represented a bit, such that `contractName` becomes a directory containing each source file that makes up the compiled contract? For example, [`contracts/LiquidGasToken.sol`](https://github.com/matnad/liquid-gas-token/blob/master/contracts/LiquidGasToken.sol) would become:

```
contracts/LiquidGasToken/LiquidGasToken.sol
contracts/LiquidGasToken/LiquidERC20.sol
contracts/LiquidGasToken/ERC20PointerSupply.sol
contracts/LiquidGasToken/SafeMath.sol
```

Another related consideration is that any private methods that aren't called are not compiled.  If `Foo` inherits `Bar` but never calls to `Bar._private_thing`, the compiler will not include `Bar._private_thing`.  Thus `Bar._private_thing` should **not** be marked as lacking coverage - it should simply be omitted.

Re: offsets vs lines - this is because Solidity denotes the end of a statement with a semi-colon rather than a new line.  And so again, making a square peg fit in a round hole here. But I've seen coverage reports for python files where a multi-line statement only considers the first line - so I think this should be possible?

brownie/_cli/test.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/python3
 
+import os
```
iamdefinitelyahuman, line 3: Would prefer to use [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html#pathlib.Path) over `os.path` to maintain consistency with the rest of the code base.

brownie/_cli/test.py
```diff
@@ -60,6 +61,13 @@ def main():
         elif isinstance(value, str):
             pytest_args.extend([opt, value])
 
+    if args['--coverage']:
+        # we want to load the coverage plugin
+        coverage_rc = os.path.abspath(
+            os.path.join(os.path.dirname(__file__), '.coveragerc'))
+        pytest_args.append('--cov=contracts')
```
iamdefinitelyahuman, line 68: The main brownie `pytest` plugin is loaded in `brownie/test/plugin.py`.  I think it would be possible to [add the dependency](https://docs.pytest.org/en/latest/plugins.html#requiring-loading-plugins-in-a-test-module-or-conftest-file) there, maybe during `pytest_configure`, without having to load `.coveragerc` this way?

It's also useful to move there because some users will invoke testing directly via `pytest` rather than `brownie test`.

brownie/network/transaction.py
```diff
@@ -278,7 +278,18 @@ def wait(self, required_confs: int) -> None:
...
+        """We run a function for every single trace so coverage could count it."""
+
     def _raise_if_reverted(self, exc: Any) -> None:
+
+        # Should check if need coverage before doing this. Also this
+        # might not be reliable indication of the end of a transaction
+        # lifecycle
+        self._expand_trace()
+        for segment in self.trace:
+            self._cover_segment(segment)
```
iamdefinitelyahuman, line 291: The actual coverage eval happens in `TransactionReceipt._expand_trace`, and the logic already involves iterating `self.trace`.  I think you could probably hook in there instead, and make it more efficient by only calling `_cover_segment` on trace steps that touch a new statement / branch.

Ignas:
Ok, almost got the contract/filename going but I am still having problems recognizing inherited function calls:
```
{'depth': 0, 'error': '', 'gas': 6700397, 'gasCost': 3, 'memory': ['0000000000000000000000000000000000000000000000000000000000000000', '0000000000000000000000000000000000000000000000000000000000000000', '0000000000000000000000000000000000000000000000000000000000000080'], 'op': 'EQ', 'pc': 439, 'stack': ['0000000000000000000000000000000000000000000000000000000070a08231', '0000000000000000000000000000000000000000000000000000000070a08231', '0000000000000000000000000000000000000000000000000000000069f03005'], 'storage': None, 'address': '0x000000000049091F98692b2460500B6D133ae31F', 'contractName': 'LiquidGasToken', 'fn': 'LiquidGasToken.balanceOf', 'jumpDepth': 0, 'source': {'filename': 'contracts/LiquidGasToken.sol', 'offset': [288, 21927]}}
```
so if balanceOf is from a baseclass - the offset pair spans the whole file that has the inheriting class... How can I reliably recognize and ignore these kinds of traces?

iamdefinitelyahuman:
I think this section is what you're looking for:

https://github.com/eth-brownie/brownie/blob/30b7eb018e55b7e504aa576e7986fdbed2371977/brownie/network/transaction.py#L600-L610

During compilation all the important instructions for identifying statements / branches are identified and flagged as such.  You can ignore any steps in the trace where `pc` doesn't have one of these flags.

More generally - all of this information is collected by functions in `brownie/tests/coverage.py`.  The final aggregated is obtained by calling [`get_merged_coverage_eval`](https://github.com/eth-brownie/brownie/blob/master/brownie/test/coverage.py#L30), which happens in the `pytest_terminal_summary` hookpoint in `brownie/tests/managers/base.py` after the test suite has finished running.  The format of this data is roughly:

```python
{
  'contractName': { 
    'path/to/source': [ 
      [statements], [branches true], [branches false]
    ]
  } 
}
```

Each of the lists is populated with numbers, which are translated to offsets based on values in `_build['coverageMap']`. An example of how this is handled can be seen in[`_build_coverage_output`](https://github.com/eth-brownie/brownie/blob/master/brownie/test/output.py#L117).

I'm not familiar with how `coverage.py` plugins work, but given all the coverage information is already being collected and aggregated, I feel like it might be simpler to just process this data after the fact in the [`pytest_terminal_summary`](https://github.com/eth-brownie/brownie/blob/master/brownie/test/managers/base.py#L217) hookpoint?

Ignas:
The hookpoint means that we lose all the help that coverage provides. The plugin handles multi processing by basically storing all tracing info in separate files for it's own database. So the rules are:

- During tracing you must output the range of lines covered by any given frame, and as that is being called during low level python tracing, it should be as cheap as possible
- You must be able to take a list of lines/things covered and produce a list of lines in a file that you want to mark as covered after the fact

This basically means that for every "trace" you have to produce - coverable identifier + range of integers (from, to), and then for every unique coverable you willl receive an aggregate of all the integers, that you have to make a report out of...

Issue #494, opened 2020-05-05
Cannot create `Contract` from etherscan contracts verified via JSON input


iamdefinitelyahuman:
### What was wrong?

`Contract.from_explorer` fails when the source code is verified using the standard JSON input.

As an example: [0xaf1610D242c7CdD30c546844aF75c147C12e94F9](https://etherscan.io/address/0xaf1610D242c7CdD30c546844aF75c147C12e94F9#code)

```python
>>> Contract.from_explorer("0xaf1610D242c7CdD30c546844aF75c147C12e94F9") 
  File "<console>", line 1, in <module>
  File "eth_brownie-1.8.2-py3.8.egg/brownie/network/contract.py", line 440, in __init__
    contract = self.from_explorer(address, owner=owner, silent=True)
  File "eth_brownie-1.8.2-py3.8.egg/brownie/network/contract.py", line 651, in from_explorer
    build = compile_and_format(
  File "eth_brownie-1.8.2-py3.8.egg/brownie/project/compiler/__init__.py", line 132, in compile_and_format
    output_json = compile_from_input_json(input_json, silent, allow_paths)
  File "eth_brownie-1.8.2-py3.8.egg/brownie/project/compiler/__init__.py", line 247, in compile_from_input_json
    return solidity.compile_from_input_json(input_json, silent, allow_paths)
  File "eth_brownie-1.8.2-py3.8.egg/brownie/project/compiler/solidity.py", line 72, in compile_from_input_json
    raise CompilerError(e)
CompilerError: Compiler returned the following errors:

Pool-flattened.sol:1:1: ParserError: Expected pragma, import directive or contract/interface/library definition.
{{
```

### How can it be fixed?
In `Contract.from_explorer`, check the format of `SourceCode` returned from Etherscan and handle accordingly.

iamdefinitelyahuman:
This issue is now causing the test suite to fail - it should be handled ASAP

PR #633, opened 2020-06-17
Feat unlock

### LINKED ISSUE #618: Add support for ganache `--unlock` flag
### Overview
Add support for the `--unlock` flag in ganache.

### Specification
* Within the network config, include`unlock`
* Automatically create `Account` objects for each listed account
* When calling `Accounts.at` in a dev environment, attempt to perform a 0 ether transfer to check if the account is unlocked. If yes, revert the tx and create the `Account` object.

matnad:
### What I did

Added the `unlock` setting to the network config in development networks.
This will pass the unlocked accounts to `ganache-cli` and add them to the `accounts` object.

Closes #618

### How I did it
Added `unlock` field to the config file, accepts a single address or multiple addresses
When `Accounts._reset()` is called, add the unlocked accounts to the end of the list.

### How to verify it
I adjusted the tests

### Checklist

### DIFF
--- a/brownie/network/account.py
+++ b/brownie/network/account.py
@@ -59,6 +59,19 @@ class Accounts(metaclass=_Singleton):
             self._accounts = [Account(i) for i in web3.eth.accounts]
         except Exception:
             pass
+
+        # Check if accounts were manually unlocked and add them
+        try:
+            unlocked_accounts = CONFIG.active_network["cmd_settings"]["unlock"]
+            if not isinstance(unlocked_accounts, list):
+                unlocked_accounts = [unlocked_accounts]
+            for address in unlocked_accounts:
+                if isinstance(address, int):
+                    address = HexBytes(address.to_bytes(20, "big")).hex()
+                self._accounts.append(Account(address))
+        except (ConnectionError, ValueError, KeyError):
+            pass
+
         if self.default not in self._accounts:
             self.default = None
 
--- a/brownie/network/account.py
+++ b/brownie/network/account.py
@@ -68,7 +68,9 @@ class Accounts(metaclass=_Singleton):
             for address in unlocked_accounts:
                 if isinstance(address, int):
                     address = HexBytes(address.to_bytes(20, "big")).hex()
-                self._accounts.append(Account(address))
+                account = Account(address)
+                if account not in self._accounts:
+                    self._accounts.append(account)
         except (ConnectionError, ValueError, KeyError):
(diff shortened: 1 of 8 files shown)


brownie/network/account.py
```diff
@@ -59,6 +59,19 @@ def _reset(self) -> None:
...
+
+        # Check if accounts were manually unlocked and add them
+        try:
+            unlocked_accounts = CONFIG.active_network["cmd_settings"]["unlock"]
+            if not isinstance(unlocked_accounts, list):
+                unlocked_accounts = [unlocked_accounts]
+            for address in unlocked_accounts:
+                if isinstance(address, int):
+                    address = HexBytes(address.to_bytes(20, "big")).hex()
+                self._accounts.append(Account(address))
```
iamdefinitelyahuman, line 71: What if the user unlocks an account that's already unlocked via their mnemonic, or includes the same account more than once in the list?

brownie/network/rpc.py
```diff
@@ -94,14 +96,22 @@ def launch(self, cmd: str, **kwargs: Dict) -> None:
...
-                cmd_list.extend([CLI_FLAGS[key], str(value)])
-            except KeyError:
-                warnings.warn(
-                    f"Ignoring invalid commandline setting for ganache-cli: "
-                    f'"{key}" with value "{value}".',
-                    InvalidArgumentWarning,
-                )
+            if key == "unlock":
+                if not isinstance(value, list):
+                    value: [value]  # type: ignore
```
iamdefinitelyahuman, line 101: should this be an `=` ?

brownie/network/rpc.py
```diff
@@ -94,14 +96,22 @@ def launch(self, cmd: str, **kwargs: Dict) -> None:
...
+                for address in value:
+                    if isinstance(address, int):
+                        address = HexBytes(address.to_bytes(20, "big")).hex()
+                    cmd_list.extend([CLI_FLAGS[key], str(address)])
```
iamdefinitelyahuman, line 105: is the conversion to `str` necessary?  I don't know if you can give any types in yaml other than `str` or `int` and you've handled integers already.

brownie/network/rpc.py
matnad: 😅

iamdefinitelyahuman:
Tests are failing here because of https://github.com/eth-brownie/brownie/issues/494

Issue #642, opened 2020-06-19
Loading a contract on Kovan results in a cloudflare error


matnad:
# What was wrong?

Running `s = SimpleStorage.at("0x123...def")` on Kovan results in a cloudflare error and invalid `403` request from the etherscan api endpoint when trying to fetch the source code.

### How can it be fixed?

We have to pass an `User-Agent` so cloudflare will not block us. This is likely also a misconfiguration from etherscan, as it only seems to happen on kovan.

PR #643, opened 2020-06-19
Fix requests header

matnad:
### What I did

Added a custom user agent to send with requests.
`{"User-Agent": f"Brownie/{__version__} (Python/{python_version})"}`
Example: `Brownie/1.9.3 (Python/3.6.8 final)"`

Related issue: #642 

### How I did it

Updated all `requests.get`  statements to include the user agent header.

### How to verify it
Try and load a contract on Kovan.

### Checklist

### DIFF
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -726,12 +726,13 @@ def _install_from_github(package_id: str) -> str:
     if install_path.exists():
         raise FileExistsError("Package is aleady installed")
 
+    headers = REQUEST_HEADERS.copy()
     if os.getenv("GITHUB_TOKEN"):
         auth = b64encode(os.environ["GITHUB_TOKEN"].encode()).decode()
-        REQUEST_HEADERS.update({"Authorization": "Basic {}".format(auth)})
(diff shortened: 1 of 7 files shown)


brownie/project/main.py
```diff
@@ -725,13 +726,12 @@ def _install_from_github(package_id: str) -> str:
     if install_path.exists():
         raise FileExistsError("Package is aleady installed")
 
-    headers: Dict = {}
     if os.getenv("GITHUB_TOKEN"):
         auth = b64encode(os.environ["GITHUB_TOKEN"].encode()).decode()
-        headers = {"Authorization": "Basic {}".format(auth)}
+        REQUEST_HEADERS.update({"Authorization": "Basic {}".format(auth)})
```
iamdefinitelyahuman, line 731: This will mutate `REQUEST_HEADERS` and leak into the calls to etherscan.  Maybe better to start by copying `REQUEST_HEADERS` to a local variable and then modifying?

matnad: 👍 Good catch, I was not aware of this problem

PR #660, opened 2020-07-04
fix: avoids development network to revert when `atexit` is being called

AlanVerbner:
### What I did

When running a script using `development` network, even using a `ganache` instance I launch, after finishing the execution the evm changes are reverted.

### How I did it

As pointed out by @iamdefinitelyahuman I removed the lines where invoked when exiting.

### How to verify it

I haven't tried myself to automatically test the `atexit` hook, I can try to do so if required (I'm not a python dev so it might take me a while but a quick googling told me that hook is only executed then the process ends, in this case, pytest) . Manual test is required I'm afraid.

### Checklist


AlanVerbner:
Happy to contribute back! the work done in this library is amazing :muscle:

Issue #719, opened 2020-08-18
Allow importing contracts inside test suites


fubuloubu:
### Overview
Brownie provides contract types as fixtures, however it would be helpful to also allow importing them via `from brownie import ContractType`, then you could do useful stuff like [this](https://docs.pytest.org/en/stable/fixture.html#parametrizing-fixtures):
```python
import pytest
from brownie import ProtocolV1, ProtocolV2

@pytest.fixture(params=[ProtocolV1, ProtocolV2])
def protocol(accounts, request):
    yield accounts[0].deploy(request.param)
```

This would parametrize the fixture such that it would run any tests it touches twice, once with `ProtocolV1`, and once with `ProtocolV2`.

iamdefinitelyahuman:
This does work in tests, but not in `conftest.py` apparently.  Shouldn't be a hard fix, just need to ensure the project is loaded before plugin collection happens.

PR #720, opened 2020-08-18
Load project before conftest

fubuloubu:
### What I did
Fixes: #719

### How I did it
Load the project config during the discovery phase of conftest.py files

### How to verify it
Everything should still work, and loading contract types in conftest.py should be possible

### Checklist

### DIFF
--- a/brownie/test/plugin.py
+++ b/brownie/test/plugin.py
@@ -56,7 +56,7 @@ def pytest_addoption(parser):
         )
 
 
-def pytest_load_initial_conftests(early_config, parser, args):
+def pytest_load_initial_conftests():
     if project.check_for_project("."):
         try:
             active_project = project.load()
@@ -107,7 +107,7 @@ def pytest_configure(config):
                 {"phases": {"explicit": True, "generate": True, "target": True}}, "brownie-failfast"
             )
 
-        active_project = project.get_loaded_projects()[-1]
+        active_project = project.get_loaded_projects()[0]
         session = Plugin(config, active_project)
         config.pluginmanager.register(session, "brownie-core")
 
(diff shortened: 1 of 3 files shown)


brownie/test/plugin.py
```diff
@@ -104,6 +107,7 @@ def pytest_configure(config):
                 {"phases": {"explicit": True, "generate": True, "target": True}}, "brownie-failfast"
             )
 
+        active_project = project.get_loaded_projects()[-1]
```
iamdefinitelyahuman, line 110: This should probably target the first list item instead of the last, in case the user loaded a 2nd project in a conftest file.

brownie/test/plugin.py
```diff
@@ -56,9 +56,8 @@ def pytest_addoption(parser):
         )
 
 
-def pytest_configure(config):
+def pytest_load_initial_conftests(early_config, parser, args):
```
iamdefinitelyahuman, line 59: If you aren't using `early_config`, `parser`, `args`, no need to include them as inputs

fubuloubu:
```suggestion
def pytest_load_initial_conftests():
```

brownie/test/plugin.py
fubuloubu:
```suggestion
        active_project = project.get_loaded_projects()[0]
```

tests/test/plugin/test_conftest.py
```diff
@@ -0,0 +1,21 @@
...
+import pytest
+from brownie import BrownieTester 
+
+@pytest.fixture(params=[BrownieTester, BrownieTester])
+def brownie_tester(request):
+    yield request.param"""
+
+
+def test_load_contract_type(plugintester):
+    plugintester.runpytest("-n 2")
```
fubuloubu, line 21:
```suggestion
    plugintester.runpytest("-n 2 --target plugin")
```

fubuloubu:
g2g

PR #734, opened 2020-08-27
Allow namespace collision between interfaces

### LINKED ISSUE #729: Name collision for interfaces in Compound finance contracts prevents `brownie` from compiling
### Environment information

* `brownie` Version: 1.10.6
* `ganache-cli` Version: n/a
* `solc` Version: 0.5.16
* Python Version: 3.7
* OS: osx

### What was wrong?

I ran:

```
git clone https://github.com/compound-finance/compound-protocol.git
```

Then in a separate directory, I first initialised brownie:

```
poetry run brownie init
```

After that, I simlinked the contracts folder:

```bash
ln -s ../compound-protocol/contracts ./contracts
```

And tried to compile:

```bash
poetry run brownie compile
```

Getting the following error:

```
  File "brownie/_cli/__main__.py", line 60, in main
    importlib.import_module(f"brownie._cli.{cmd}").main()
  File "brownie/_cli/compile.py", line 37, in main
    project.load()
  File "brownie/project/main.py", line 656, in load
    return Project(name, project_path)
  File "brownie/project/main.py", line 160, in __init__
    self.load()
  File "brownie/project/main.py", line 170, in load
    self._sources = Sources(contract_sources, interface_sources)
  File "brownie/project/sources.py", line 53, in __init__
    + "\n  ".join(f"{k}: {', '.join(sorted(v))}" for k, v in collisions.items())
NamespaceCollision: Multiple contracts or interfaces with the same name
  PotLike: contracts/CDaiDelegate.sol, contracts/DAIInterestRateModelV3.sol
```

### How can it be fixed?

It seems like `brownie` doesn't support contracts with the same names, I wonder if it should be more permissive, perhaps namespacing contracts that appear in different files by prefixing with the filename (certainly for interfaces). In the `compound` example, these interfaces actually have different source code and are meant to be deployed as part of another contract.

iamdefinitelyahuman:
### What I did
Lessen the namespace collision restrictions.

Closes #729 

### How I did it
When looking for collisions, allow multiple uses of the same name so long as only one of the objects is a contract. This way projects using interfaces with the same name as the actual contract will now compile.

### How to verify it
Try to compile https://github.com/compound-finance/compound-protocol using Brownie

### DIFF
--- a/brownie/project/sources.py
+++ b/brownie/project/sources.py
@@ -22,28 +22,33 @@ class Sources:
         self._contracts: Dict = {}
         self._interfaces: Dict = {}
 
+        contracts: Dict = {}
         collisions: Dict = {}
         for path, source in contract_sources.items():
             self._source[path] = source
             if Path(path).suffix != ".sol":
-                contract_names = [Path(path).stem]
+                contract_names = [(Path(path).stem, "contract")]
             else:
                 contract_names = get_contract_names(source)
-            for name in contract_names:
-                if name in self._contracts:
-                    collisions.setdefault(name, set()).update([path, self._contracts[name]])
-                self._contracts[name] = path
+            for name, type_ in contract_names:
+                if name in contracts:
+                    if type_ == "interface":
+                        # allow names to overlap when dealing with interfaces
+                        continue
+                    if contracts[name][1] != "interface":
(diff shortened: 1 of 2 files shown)


priyankaj1910:
I am getting this error

PR #826, opened 2020-10-28
fix: filter heuristics from dev revert map

iamdefinitelyahuman:
### What I did
When generating the dev revert map, do not include revert strings generated from heuristics.

Fixes an issue encountered with Curve DAO's test suite: https://github.com/curvefi/curve-dao-contracts/runs/1317870437?check_suite_focus=true#step:8:1361

### How I did it
Filter revert messages that do not begin with `dev: `

### DIFF
--- a/brownie/project/build.py
+++ b/brownie/project/build.py
@@ -88,12 +88,15 @@ class Build:
                 except (KeyError, ValueError):
                     pass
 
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman:
This ended up being the wrong approach.

PR #846, opened 2020-11-16
Allow replacing pending transactions

### LINKED ISSUE #726: Allow manual repricing of pending transactions
### Overview
It should be simple to reprice a pending transaction.

### Specification
In the console, with a `TransactionReceipt` object that's still pending:

```
>>> tx.reprice(gas_price)
```

Once this is done, the logic in `TransactionReceipt._await_confirmation` must watch both pending transactions - it could still be possible that the original one confirms before the higher-priced one is seen by the miner.

### Dependencies
None.

iamdefinitelyahuman:
### What I did
Add the `TransactionReceipt.replace` method for replacing pending transactions.

Closes #726

### How I did it
The logic flow for creating a `TransactionReceipt` had to be modified slightly. Creating the object is now optionally-blocking via the `is_blocking` kwarg (this had to be done to avoid making a breaking change). When broadcasting, the object is initially created via a non-blocking call, then added to `history`, then if `required_confs` is non-zero it calls `wait` to block until the necessary confirmations. By using this approach, each tx object is added to history immediately, so that users can `CTRL-C` and still access the pending the tx.

Replacement happens via `TransactionReceipt.replace` which accepts either an `increment` modifier or `gas_price` absolute and returns a new tx object.

As soon as any one transaction confirms, the other tx's are marked with status `-2` (dropped) and removed from `history`.

### How to verify it
This is difficult to write tests for.  I've played with it a fair bit on the mainnet and everything seems to work, but unfortunately I think the only real test here is to release and solve bugs are they appear.

### DIFF
--- a/brownie/network/transaction.py
+++ b/brownie/network/transaction.py
@@ -26,7 +26,7 @@ from brownie.utils import color
 from brownie.utils.output import build_tree
 
 from . import state
-from .event import _decode_logs, _decode_trace
+from .event import EventDict, _decode_logs, _decode_trace
 from .web3 import web3
 
 
@@ -114,6 +114,7 @@ class TransactionReceipt:
         sender: Any = None,
         silent: bool = True,
         required_confs: int = 1,
+        is_blocking: bool = True,
         name: str = "",
         revert_data: Optional[Tuple] = None,
     ) -> None:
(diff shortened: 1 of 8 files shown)


brownie/network/transaction.py
```diff
@@ -58,6 +59,13 @@ def wrapper(self: "TransactionReceipt", *args: Any, **kwargs: Any) -> Any:
     return wrapper
 
 
+class Status(IntEnum):
+    Dropped = -2
+    Pending = -1
+    Reverted = 0
```
fubuloubu, line 65: I feel like these two should be swapped?

iamdefinitelyahuman: 0 and 1 are standard values to represent reverted and successful in a txreceipt. Brownie has used -1 for pending since forever, modifying it would be a breaking change :grimacing:

PR #847, opened 2020-11-17
Automatic transaction repricing / replacement

### LINKED ISSUE #727: Automatic transaction repricing
### Overview
Automatically rebroadcast pending transactions with a higher gas price, until the transaction confirms.

### Specification
Not 100% what's best, but I'm thinking:

``` MyContract.foo({'gas_price': "50 gwei", 'gas_price_increment': "1 gwei", 'gas_price_interval': 60})```
 
`gas_price_increment`: the amount of gas to add with each subsequently broadcasted tx. Including this value enables the functionality.
`gas_price_interval`: the frequency to rebroadcast at.  If not given, should use a sensible default (1 minute seems good)

`pymaker` already [does this well](https://github.com/makerdao/pymaker/blob/master/pymaker/__init__.py#L525) - it would be good to look at their implementation for inspiration and solutions to problems I've not yet considered.

### Dependencies
#726

### PR DESCRIPTION
**Pull request by iamdefinitelyahuman**
### What I did
Automatic repricing of transactions.

Closes #727 

### How I did it
When setting `gas_price`, you may now use an object that subclasses from `GasABC`:

* `SimpleGasStrategy` for simple strategies that are called once
* `BlockGasStrategy` which is called every `n` blocks
* `TimeGasStrategy` which is called every `n` seconds

For strategies that trigger tx replacement, returning a non-`None` value broadcasts a new transaction.

Thus far I have created two actual strategies, based around the [gasnow](https://www.gasnow.org/) API.

### How to verify it
Try it out!

### TODO

- [x] more strategies
- [x] consideration / updates to logic re: console output while waiting for a tx to confirm
- [x] detailed natspec for strategies and ABCs
- [x] documentation
- [ ] unit testing (not sure how this will look)

### DIFF
--- a/brownie/network/gas/strategies.py
+++ b/brownie/network/gas/strategies.py
@@ -1,0 +1,59 @@
+import threading
+import time
+
+import requests
+
+from .bases import BlockGasStrategy, SimpleGasStrategy
+
+_gasnow = {"time": 0, "data": None}
+_gasnow_lock = threading.Lock()
+
+
+def _fetch_gasnow(key):
+    with _gasnow_lock:
+        if time.time() - _gasnow["time"] > 15:
+            data = None
+            for i in range(12):
+                response = requests.get(
+                    "https://www.gasnow.org/api/v3/gas/price?utm_source=brownie"
+                )
+                if response.status_code != 200:
+                    time.sleep(5)
+                    continue
+                data = response.json()["data"]
+            if data is None:
+                raise ValueError
+            _gasnow["time"] = data.pop("timestamp") // 1000
+            _gasnow["data"] = data
+
+    return _gasnow["data"][key]
+
+
+class GasNowStrategy(SimpleGasStrategy):
+    def __init__(self, speed: str = "fast"):
+        if speed not in ("rapid", "fast", "standard", "slow"):
+            raise ValueError("`speed` must be one of: rapid, fast, standard, slow")
+        self.speed = speed
+
+    def get_gas_price(self):
+        return _fetch_gasnow(self.speed)
+
+
+class GasNowScalingStrategy(BlockGasStrategy):
+    def __init__(
+        self, initial_speed: str = "standard", increment: float = 1.1, block_duration: int = 2
+    ):
+        super().__init__(block_duration)
+        if initial_speed not in ("rapid", "fast", "standard", "slow"):
+            raise ValueError("`initial_speed` must be one of: rapid, fast, standard, slow")
+        self.speed = initial_speed
+        self.increment = increment
+
+    def get_gas_price(self, current_gas_price, elapsed_blocks):
+        if current_gas_price is None:
+            return _fetch_gasnow(self.speed)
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by iamdefinitelyahuman** [comment]
Still need to add another strategy or two here, but this is close!

cc @banteg @fubuloubu

File: brownie/network/gas/strategies.py [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+        self.initial_speed = initial_speed
+        self.max_speed = max_speed
+        self.increment = increment
+
+    def update_gas_price(self, last_gas_price: int, elapsed_blocks: int) -> Optional[int]:
+        max_gas_price = _fetch_gasnow(self.max_speed)
+        new_gas_price = max(int(last_gas_price * self.increment), _fetch_gasnow(self.initial_speed))
+        if new_gas_price <= max_gas_price:
+            return new_gas_price
+        return None
```
fubuloubu, line 91: This seems like difficult behavior to require, and `None` is a bit magical.

Instead, why can't the strategy always return a value, and a rebroadcast is only triggered by the handler if `new_gas_price <= max_gas_price`. This way that logic doesn't need to be replicated across every single strategy.

It simply doesn't make sense to rebroadcast a lower price, because that'll never confirm over one with a higher price.

fubuloubu: Also, I still think something with iterators would work really nice here, e.g.:

```
gas_price = _fetch_gasnow(self.initial_speed)
while gas_price < _fetch_gasnow(self.max_speed):
    gas_price = int(_fetch_gasnow(self.initial_speed) * self.increment)
    yield gas_price
```

banteg: A new gas price should only be yielded if the increase is sufficient. Geth nodes accept [at least 10%](https://github.com/ethereum/go-ethereum/blob/706f5e3b98a419e8b93653d21876b74a60ab3ee6/core/tx_pool.go#L160) bump, Openethereum (Parity) accepts [at least 12.5%](https://github.com/openethereum/openethereum/blob/27a0142af14730bcb50eeacc84043dc6f49395e8/miner/src/pool/scoring.rs#L38).

banteg:
Another potentially useful default strategy for Geth with graphql enabled:
```python
def gas_price(position=500):
    query = "{ pending { transactions { gasPrice }}}"
    resp = requests.post(f"{ETH_RPC_URL}/graphql", json={"query": query})
    data = resp.json()["data"]["pending"]["transactions"]
    prices = [int(x["gasPrice"], 16) for x in data]
    return Wei(sorted(prices, reverse=True)[:position][-1])
```

brownie/network/gas/strategies.py
iamdefinitelyahuman: Iterators implemented!

banteg (dismissed):

iamdefinitelyahuman:
This should be gtg now

banteg (approved):

fubuloubu (approved):
Gave it a shot, looks pretty good to me.

Issue #765, opened 2020-09-19
Allow broadcasting of reverting transactions in live environment


iamdefinitelyahuman:
### Overview
It should be possible to broadcast a transaction that will revert.

### Specification
Currently, would-be reverting transactions in a live environment are blocked twice:

* if a gas limit is not set, the call to `eth_estimateGas` will revert.
* if `reverting_tx_gas_limit` is not set in the config, `Account._check_for_reverting_tx` is called. This uses `eth_call` to verify that a tx will not revert prior to broadcasting.

Usually this is a good thing, but in some cases (such as simulating a failing tx on a testnet) it's not desirable.  It should be possible to disable this behavior, either by a config setting or a parameter in the transaction dict such as `{"allow_revert": True}`.

In cases where a transaction is blocked by either of these actions, the error message should be improved to explain what has happened and how it can be circumvented.

PR #852, opened 2020-11-21
Fix: TransactionReceipt for externally broadcasted tx's

iamdefinitelyahuman:
### What I did
Fix two issues when attempting to create `TransactionReceipt` objects for externally broadcasted tx's:

* `EthAddress` object has no nonce, so querying the nonce is now handled via `web3.eth.getTransactionCount`
* even when a transaction request is non-blocking, wait for the transaction receipt if the transaction shows as already confirmed. This fixes an issue where `chain.get_transaction` was always returning tx's marked as pending (that were then marked as confirmed half a second later).

### How to verify it
Run tests.  I added some cases to query txid objects from the mainnet that were failing prior to this PR.

### DIFF
--- a/brownie/network/transaction.py
+++ b/brownie/network/transaction.py
@@ -310,10 +310,12 @@ class TransactionReceipt:
                 tx: Dict = web3.eth.getTransaction(self.txid)
                 break
             except TransactionNotFound:
-                if self.sender.nonce > self.nonce:  # type: ignore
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman:
This was originally handled as part of #847, but is also required for #765 (which is a blocker to solving a bug in #847) so I've moved it to a separate PR.

Issue #411, opened 2020-04-14
Allow publishing source on Etherscan


iamdefinitelyahuman:
### Overview
It should be possible (and simple) to verify a deployed contract's source code on etherscan.

### Specification
Not 100% on how this would be best implemented. Some thoughts:

1. Add an internal "flattener" to turn multiple source files into a single file. Might also be worth exposing this on it's own via the CLI.
2. Add a method to `Contract`, to be called from in the console or a script.
3. Consider a flag in `brownie run` to auto-verify all sources after deployment?

Possible syntax:

```python
contract = ContractContainer.deploy({'from': accounts[0]})
contract.publish_on_etherscan()
```
and / or
```python
ContractContainer.deploy({'from': accounts[0]}, publish=True)
```

Skyge:
Does anyone work on this issue? if not, I would like to solve this issue.

WyseNynja:
Etherscan updated to support verifying from solidity-standard-json-input.

The `publish_on_etherscan` function should send [a POST like this](https://etherscan.io/sourcecode-demo.html).

Netherdrake:
What is the current best way to achieve this for contracts created trough `brownie`?

fubuloubu:
I think it would be good practice to have a `publish_sources: True|False` key that pushes to etherscan (or whatever block explorer) whenever a contract is deployed to a public network.

sabotagebeats:
The json seems located in `build/contracts/` but where can I find the hex encoded constructor abi within `build/deployments/` ? 

[https://github.com/eth-brownie/brownie/issues/766#issuecomment-711619114](url)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.5 ETH (236.29 USD @ $472.58/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/eth-brownie/brownie/411/100024068).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/eth-brownie/brownie/411/100024068).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://chat.gitcoin.co/'>Gitcoin Chat</a>
* $1,045,515.92 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

[...]

PR #914, opened 2020-12-28
Feat publish source

matnad:
### What I did

Added a method to publish source to etherscan for solidity contracts.
`DeployedContract.publish_source()`

Related issue: #411 

### How I did it
Flatten source code into one file (thanks Ben for starting on this). Make sure to include all comments, proper licensing and exactly one pragma statement. 
Will wait for contract to be recognized by etherscan after deploying, then submit request for verification and finally wait for the verification to be processed and report the result.

### How to verify it
`deployed = MyContract.deploy({'from': account}, publish_source=True)`
and also 
`MyContract.get_verification_info()`
image
-> https://ropsten.etherscan.io/address/0x6De9c6197bB8f61177C30946d90CAC2720272c5c#code

Tests are adapted for the changes to the solc versioning, but no new tests for this feature are written yet.

### Checklist

### DIFF
--- a/brownie/network/contract.py
+++ b/brownie/network/contract.py
@@ -5,6 +5,7 @@ import os
 import re
 import time
 import warnings
+from collections import defaultdict
 from pathlib import Path
 from textwrap import TextWrapper
 from typing import Any, Dict, Iterator, List, Optional, Set, Tuple, Union
@@ -38,6 +39,7 @@ from brownie.exceptions import (
 from brownie.project import compiler, ethpm
 from brownie.typing import AccountsType, TransactionReceiptType
 from brownie.utils import color
+from brownie.utils.toposort import toposort_flatten
 
 from . import accounts, chain
 from .event import _add_deployment_topics, _get_topics
@@ -142,27 +144,43 @@ class _ContractBase:
         elif language == "Solidity":
             flattened_source = ""
             has_abiencoder = False
-            abiencoder_str = "\npragma experimental ABIEncoderV2;"
             abiencoder_re = r"(^|\r|\n|\r\n)\s*pragma experimental ABIEncoderV2;"
+
+            dependency_tree = defaultdict(set)
+            dependency_tree["__root_node__"] = set(self._build["dependencies"])
+            code_fragments = {}
+
+            used_offsets = defaultdict(list)  # aggregate all used offsets per source file
 
             for name in self._build["dependencies"]:
                 build_json = self._project._build.get(name)
+                if "dependencies" in build_json:
+                    dependency_tree[name].update(build_json["dependencies"])
                 offset = slice(*build_json["offset"])
+                used_offsets[build_json["source"]].append((offset.start, offset.stop))
                 source = self._slice_source(build_json["source"], offset)
                 if "sourcePath" in build_json:
-                    file_name = Path(build_json["sourcePath"]).parts[-1]
+                    part_name = Path(build_json["sourcePath"]).parts[-1]
                 else:
-                    file_name = f"{build_json['contractName']}.sol"
-                flattened_source = f"{flattened_source}\n\n// File: {file_name}\n\n{source}"
+                    part_name = f"{build_json['contractName']}"
+                code_fragments[name] = [source, part_name]
                 if not has_abiencoder:
                     has_abiencoder = (
                         len(re.findall(abiencoder_re, build_json["source"][: offset.start])) > 0
                     )
 
+            for name in toposort_flatten(dependency_tree):
+                if name == "__root_node__":
+                    continue
+                source, part_name = code_fragments[name]
+                flattened_source = f"{flattened_source}\n\n// Part: {part_name}\n\n{source}"
+
+            # Top level contract, defines compiler and license
             build_json = self._build
             version = build_json["compiler"]["version"]
             version_short = re.findall(r"^[^+]+", version)[0]
             offset = slice(*build_json["offset"])
+            used_offsets[build_json["source"]].append((offset.start, offset.stop))
             source = self._slice_source(build_json["source"], offset)
             file_name = Path(build_json["sourcePath"]).parts[-1]
             licenses = re.findall(
@@ -173,10 +191,31 @@ class _ContractBase:
                 has_abiencoder = (
                     len(re.findall(abiencoder_re, build_json["source"][: offset.start])) > 0
                 )
+
+            # find global structs and enums
+            global_structs = set()
+            global_enums = set()
+            for src, offsets in used_offsets.items():
+                offsets = sorted(offsets)
+                offsets.append((len(src), len(src)))
+                start = 0
+                for offset in offsets:
+                    cleaned_section = remove_comments(src[start:offset[0]])
+                    global_structs.update(re.findall(r"(\w*struct [^}]+})", cleaned_section))
+                    global_enums.update(re.findall(r"(\w*enum [^}]+})", cleaned_section))
+                    start = offset[1]
+
+            # combine to final flattened source
+            lb = "\n"
+            abiencoder_str = "\npragma experimental ABIEncoderV2;"
+            is_global = len(global_enums) + len(global_structs) > 0
+            global_str = "// Global Enums and Structs\n\n" if is_global else ""
             flattened_source = (
                 f"// SPDX-License-Identifier: {license_identifier}\n\n"
                 f"pragma solidity {version_short};"
-                f"{abiencoder_str if has_abiencoder else ''}{flattened_source}\n\n"
+                f"{abiencoder_str if has_abiencoder else ''}\n\n{global_str}"
+                f"{lb.join(global_enums)}\n\n{lb.join(global_structs)}"
+                f"{flattened_source}\n\n"
                 f"// File: {file_name}\n\n{source}\n"
             )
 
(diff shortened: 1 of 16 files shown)


matnad:
@Macarse  @banteg feel free to test the branch and provide feedback!

brownie/network/contract.py
```diff
@@ -530,6 +591,136 @@ def balance(self) -> Wei:
...
+
+        payload: Dict = {
+            "apikey": api_key,
+            "module": "contract",
+            "action": "verifysourcecode",
+            "contractaddress": address,
+            "sourceCode": flat["flattened_source"],
+            "codeformat": "solidity-single-file",
+            "contractname": flat["contract_name"],
+            "compilerversion": "".join(("v", flat["compiler_version"])),
```
iamdefinitelyahuman, line 666: What is the reason you didn't use an f-string here?

brownie/network/contract.py
```diff
@@ -127,6 +128,66 @@ def decode_input(self, calldata: Union[str, bytes]) -> Tuple[str, Any]:
 
         return function_sig, input_args
 
+    def flatten(self) -> dict:
```
iamdefinitelyahuman, line 131:
```suggestion
    def flatten(self) -> Dict:
```

brownie/network/contract.py
```diff
@@ -127,6 +128,66 @@ def decode_input(self, calldata: Union[str, bytes]) -> Tuple[str, Any]:
...
+            licenses = re.findall(
+                r"SPDX-License-Identifier:(.*)\n", build_json["source"][: offset.start]
+            )
+            license_identifier = licenses[0].strip() if len(licenses) == 1 else "NONE"
+            flattened_source = (
+                f"// SPDX-License-Identifier: {license_identifier}\n\n"
+                f"pragma solidity {version_short};{flattened_source}\n\n"
+                f"// File: {file_name}\n\n{source}\n"
+            )
+            return {
```
iamdefinitelyahuman, line 164: I understand the importance of returning all this information for source verification, but with Vyper the function is only returning a source file.  In order to keep the return value consistent it might be better to simply raise when language == Vyper and say "you don't need to flatten this".  ?

iamdefinitelyahuman (approved):
:rocket:

brownie/network/contract.py
matnad: Agreed, as long as vyper verification is not implemented via API, we should raise.

matnad:
Update:
- requested changes
- Handle `pragma experimental ABIEncoderV2`
- Auto-fetch correct constructor arguments
- Temporary fix for interfaces being saved into `build/contracts/` (will discuss with @iamdefinitelyahuman for a better solution)

TODO:
- [ ]  Ensure correct order of files in the flattened file (base before derived is a hard requirement)
- [ ] Fix interface/contract issue

Macarse (dismissed):

Macarse:
@iamdefinitelyahuman I have tested the last version: https://github.com/eth-brownie/brownie/pull/914/commits/763a5c827987a21632dac9d238d7d0afe4da4d96
 and I could verify a couple of contracts I couldn't verify in the past. I believe this is a good first version for contract verification 🚀 
 
 Great work @matnad h/t

matnad:
This turned out to be a monster of a feature! A lot of edge cases have been handled and I currently can't find a repo that doesn't verify (including all brownie mixes). **Happy to merge after a review.**

For repos that don't work please open an issue and I will look at it.

brownie/network/contract.py
```diff
@@ -127,6 +130,153 @@ def decode_input(self, calldata: Union[str, bytes]) -> Tuple[str, Any]:
...
+        language = self._build["language"]
+        if language == "Vyper":
+            raise TypeError(
+                "Etherscan does not support API verification of source code "
+                "for vyper contracts. You need to verify the source manually"
+            )
+        elif language == "Solidity":
+            flattened_source = ""
+            has_abiencoder = False
+            abiencoder_re = r"(^|\r|\n|\r\n)\s*pragma experimental ABIEncoderV2;"
```
Macarse, line 147: Should we move all this parsing logic into a separate class?

docs/api-network.rst
```diff
@@ -297,7 +297,7 @@ Account Methods
         >>> accounts[0].balance() == "100 ether"
         True
 
-.. py:classmethod:: Account.deploy(contract, *args, amount=None, gas_limit=None, gas_price=None, nonce=None, required_confs=1)
+.. py:classmethod:: Account.deploy(contract, *args, amount=None, gas_limit=None, gas_price=None, nonce=None, required_confs=1, allow_revert=False, silent=False, publish_source=False,)
```
Macarse, line 300: I wonder if `silent` should be `True ` by default. I would like to avoid the print of `Verification complete. Result: Pass - Verified` for end users.

tests/project/compiler/test_solidity.py
```diff
@@ -58,11 +58,11 @@ def msolc(monkeypatch):
 
 def test_set_solc_version():
     compiler.set_solc_version("0.5.7")
-    assert solcx.get_solc_version().truncate() == compiler.solidity.get_version()
-    assert solcx.get_solc_version().truncate() == Version("0.5.7")
+    assert solcx.get_solc_version(True) == compiler.solidity.get_version()
```
Macarse, line 61: no idea what the boolean is used of the `get_solc_version()` call but it looks like `True` is used in every instance in this PR. Shouldn't that be the default?

Macarse (approved):

tests/project/compiler/test_solidity.py
matnad: It will be the default moving on, we had to make some compiler changes and to not break anything we added `True` to use the updated version. Once everything is changed and working, we can move to the new version entirely.

brownie/network/contract.py
matnad: That is Ben's call :P
The contract.py needs to be refactored anyways at some point... I don't think we should do that in this PR.

docs/api-network.rst
matnad: I think some feedback is nice, especially for longer calls and for newer users. Advanced users will know how to disable it. 
@iamdefinitelyahuman you have an opinion on this?

iamdefinitelyahuman: The `silent` kwarg is applying to both verification and deployment..  I don't think we'd want to silence the output on the actual deployment action, so unless we split it into 2 silent kwargs, I vote it's `False` by default.

brownie/network/contract.py
iamdefinitelyahuman: This probably should only be available on the `ContractContainer`.  Otherwise someone will complain when their contract has a `get_verification_info` function and can't be used with brownie cuz of namespace collision.

brownie/network/contract.py
```diff
@@ -127,6 +130,153 @@ def decode_input(self, calldata: Union[str, bytes]) -> Tuple[str, Any]:
...
+                    )
+
+            for name in toposort_flatten(dependency_tree):
+                if name == "__root_node__":
+                    continue
+                source, part_name = code_fragments[name]
+                # Create a duplicate entry for each import alias
+                part_name_short = part_name
+                if "/" in part_name:
+                    part_name_short = part_name.split("/")[-1]
```
iamdefinitelyahuman, line 186: We know for sure this is OK on windows, yes? Solidity always uses unix-style paths?

brownie/network/contract.py
```diff
@@ -127,6 +130,153 @@ def decode_input(self, calldata: Union[str, bytes]) -> Tuple[str, Any]:
...
+            # find global structs and enums
+            global_structs = set()
+            global_enums = set()
+            for src, offsets in used_offsets.items():
+                offsets = sorted(offsets)
+                offsets.append((len(src), len(src)))
+                start = 0
+                for o in offsets:
+                    cleaned_section = remove_comments(src[start : o[0]])
+                    global_structs.update(re.findall(r"(\w*struct [^}]+})", cleaned_section))
```
iamdefinitelyahuman, line 224: It would be better to find these using the AST.  Regex is too prone to false-positives...  someone might have a commented out struct or enum.

brownie/network/contract.py
```diff
@@ -530,6 +684,151 @@ def balance(self) -> Wei:
         balance = web3.eth.getBalance(self.address)
         return Wei(balance)
 
+    def publish_source(self, silent: bool = False) -> bool:
```
iamdefinitelyahuman, line 687: Same comment as above...  this is probably better as `ContractContainer.publish_source(contract)` where you hand it a `Contract` object.  Otherwise more collisions :scream:

tests/project/compiler/test_solidity.py
iamdefinitelyahuman: For readability, suppose it could say `with_commit_hash=True`

iamdefinitelyahuman (commented):
Coming along nicely..

docs/api-network.rst
matnad: I think two args is overkill for this. Since verification takes around 30 secs+, I think it is nice to let the user know why it is stalling.

brownie/network/contract.py
matnad: Yes, this is taken from the build jsons, the path is always with /.

matnad:
Big update.
- Complete refactor of `get_verification_info()` to use py-solc-ast. No more regex (except one that is safe). This is much cleaner now!
- Moved the methods to `ContractContainer` and adjusted docs.
- Added a simple test to check if verification info breaks.
- Tested functionality with various repos (brownie mixes, private repos, testing repos).

brownie/network/contract.py
```diff
@@ -239,6 +243,309 @@ def _add_from_tx(self, tx: TransactionReceiptType) -> None:
...
+
+            pragma_statements = set()
+            global_structs = set()
+            global_enums = set()
+            import_aliases: Dict = defaultdict(list)
+            for n, src in [ns.values() for ns in nodes_source]:
+                pragmas = n.children(filters={"nodeType": "PragmaDirective"})
+                for pragma in pragmas:
+                    src_offsets = [int(x) for x in pragma.get("src").split(":")]
+                    pragma_statements.add(src[src_offsets[0] : src_offsets[0] + src_offsets[1]])
```
iamdefinitelyahuman, line 277:
```suggestion
                    pragma_statements.add(src[slice(*pragma.offset)])
```

`solcast` node objects already expose the offset formatted in the way you require.

iamdefinitelyahuman: I think this change can be applied in a few more places below

iamdefinitelyahuman (approved):
Amazing stuff..  this clearly has been a huge amount of work.  The community loves you :heart:

brownie/network/contract.py
matnad: Good catch, thanks!

iamdefinitelyahuman (approved):

misterblonde:
I still have the issue, even when I'm using `publish_source=True` and your new `get_verification_info()` afterwards. I only have one solidity script with one compiler version, so that shouldn't be the issue. Has this been continued elsewhere/

turkishfilms:
im getting a jsondecodererror when i use publish_source=True. I was able to manually verify(copy and paste) but wasnt able to automatically verify. only one .sol file

Issue #915, opened 2020-12-30
Brownie mixes are assumed to use "master" as primary branch


iamdefinitelyahuman:
### What was wrong?
`brownie bake` assumes that the primary branch of a repository is named "master". Github's recent change of the default branch name to "main" means that this will inevitably create an issue for new mixes.

### How can it be fixed?
Probably there is an API endpoint that can be called to check for the name of the primary branch on a repository. This should be called first, prior to attempting to download a mix, so that any branch name will work.

PR #917, opened 2021-01-01
fix: make bake cmd use default brownie-mix branch

skellet0r:
Previously, `brownie bake` would assume the master branch was in each
brownie-mix repository. Now it checks the github api to verify the
default branch first.

Resolves eth-brownie/brownie#915

### What I did

- Modified `MIXES_URL` to now require 2 arguments when formatting the string, the project_name and default_branch
- Created a function `_get_mix_default_branch(mix_name: str) -> str`, which given the brownie-mix name, returns the default branch for the repository
- Modified `from_brownie_mix` function to call `_get_mix_default_branch`, and include the returned value when formatting `MIXES_URL`

Related issue: #915 

### How I did it

I checked [stackoverflow](https://stackoverflow.com/questions/16500461/how-do-i-find-the-default-branch-for-a-repository-using-the-github-v3-api) for how to find the default branch of a repository. Using the [github api docs](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-a-repository) and `curl`, I verified that endpoint worked and that `default_branch` was a key in the json returned. I then created the function `_get_mix_default_branch` to encapsulate the github api request, and then implemented it in the `from_brownie_mix` function.  

### How to verify it

Currently, there are no brownie-mix repositories with a default branch not named `master`. However, the implementation does not break `tests/project/test_brownie_mix.py`.

### Checklist

### DIFF
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -974,7 +974,7 @@ def _get_mix_default_branch(mix_name: str) -> str:
     str
         The default branch name on github.
     """
-    REPO_GH_API = "https://api.github.com/repos/brownie-mix/{}-mix".format(mix_name)
+    REPO_GH_API = f"https://api.github.com/repos/brownie-mix/{mix_name}-mix"
     r = requests.get(REPO_GH_API, headers=REQUEST_HEADERS)
     if r.status_code != 200:
         status, repo, message = r.status_code, f"brownie-mix/{mix_name}", r.json()["message"]
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -53,7 +53,7 @@ from brownie.project.sources import Sources, get_pragma_spec
 from brownie.utils import notify
 
 BUILD_FOLDERS = ["contracts", "deployments", "interfaces"]
-MIXES_URL = "https://github.com/brownie-mix/{}-mix/archive/master.zip"
+MIXES_URL = "https://github.com/brownie-mix/{}-mix/archive/{}.zip"
 
 GITIGNORE = """__pycache__
 .history
@@ -588,7 +588,8 @@ def from_brownie_mix(
     Returns the path to the project as a string.
     """
     project_name = str(project_name).lower().replace("-mix", "")
-    url = MIXES_URL.format(project_name)
+    default_branch = _get_mix_default_branch(project_name)
+    url = MIXES_URL.format(project_name, default_branch)
     if project_path is None:
         project_path = Path(".").joinpath(project_name)
     project_path = Path(project_path).resolve()
@@ -958,3 +959,21 @@ def _stream_download(download_url: str, target_path: str) -> None:
 
     with zipfile.ZipFile(BytesIO(content)) as zf:
         zf.extractall(target_path)
+
+
+def _get_mix_default_branch(mix_name: str) -> str:
+    """Get the default branch for a brownie-mix repository.
+
+    Arguments
+    ---------
+    mix_name : str
+        Name of a brownie-mix repository without -mix appended.
+
+    Returns
+    -------
+    str
+        The default branch name on github.
+    """
+    REPO_GH_API = "https://api.github.com/repos/brownie-mix/{}-mix".format(mix_name)
+    r = requests.get(REPO_GH_API).json()
+    return r["default_branch"]
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -989,7 +989,7 @@ def _get_mix_default_branch(mix_name: str) -> str:
             msg = "".join(msg_lines)
         raise ConnectionError(msg)
     elif "default_branch" not in r.json():
-        msg = "API results did not include {}'s default branch".format(mix_name)
+        msg = f"API results did not include {mix_name}'s default branch"
         raise KeyError(msg)
 
     return r.json()["default_branch"]
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -975,5 +975,20 @@ def _get_mix_default_branch(mix_name: str) -> str:
         The default branch name on github.
     """
     REPO_GH_API = "https://api.github.com/repos/brownie-mix/{}-mix".format(mix_name)
-    r = requests.get(REPO_GH_API).json()
-    return r["default_branch"]
+    r = requests.get(REPO_GH_API, headers=REQUEST_HEADERS)
+    if r.status_code != 200:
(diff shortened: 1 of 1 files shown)


brownie/project/main.py
```diff
@@ -958,3 +959,21 @@ def _stream_download(download_url: str, target_path: str) -> None:
...
+    mix_name : str
+        Name of a brownie-mix repository without -mix appended.
+
+    Returns
+    -------
+    str
+        The default branch name on github.
+    """
+    REPO_GH_API = "https://api.github.com/repos/brownie-mix/{}-mix".format(mix_name)
+    r = requests.get(REPO_GH_API).json()
```
matnad, line 978: We use custom headers for requests made with brownie, they are available via `REQUEST_HEADERS` and should be added to this request (also see other requests in this file).

You should also check if the request response is valid and contains what you need and throw with a more meaningful error otherwise. We generally check `status_code` and if the required data field exists. You can look at other requests in the file to get an Idea.

brownie/project/main.py
```diff
@@ -588,7 +588,8 @@ def from_brownie_mix(
     Returns the path to the project as a string.
     """
     project_name = str(project_name).lower().replace("-mix", "")
-    url = MIXES_URL.format(project_name)
+    default_branch = _get_mix_default_branch(project_name)
+    url = MIXES_URL.format(project_name, default_branch)
```
matnad, line 592: Line 601: ` project_path.parent.joinpath(project_name + "-mix-master").rename(project_path)`

Here the reference to master needs to be changed too.

matnad (changes requested):
Looking good so far, thanks!

brownie/project/main.py
skellet0r: @matnad resolved now with cfc73a1. Thanks for the pointer!

skellet0r:
@matnad Thank you for those reviews, very good catches. Definitely will have to get more acquainted with the code base and customs when contributing. I made those changes with the commits cfc73a1 and 2e16fa5.

brownie/project/main.py
skellet0r: Fixed with 2e16fa5

brownie/project/main.py
```diff
@@ -597,7 +598,9 @@ def from_brownie_mix(
 
     print(f"Downloading from {url}...")
     _stream_download(url, str(project_path.parent))
-    project_path.parent.joinpath(project_name + "-mix-master").rename(project_path)
+    project_path.parent.joinpath(project_name + "-mix-{}".format(default_branch)).rename(
+        project_path
+    )
```
matnad, line 603: Style wise we generally use f-strings over `.format()` wherever possible and only use the latter when the string needs to be changed at a later point (as in line 56 and then 592). A simple string such as here should be written as f-string like `f"{project_name}-mix-{default_branch}"`. It helps with readability.

There are more strings in the PR further down that should also be changed from `.format()` to f-strings.

brownie/project/main.py
```diff
@@ -958,3 +961,36 @@ def _stream_download(download_url: str, target_path: str) -> None:
...
+    if r.status_code != 200:
+        msg = "Status {} when retrieving repo brownie-mix/{} information from GHAPI: '{}'".format(
+            r.status_code, mix_name, r.json()["message"]
+        )
+        if r.status_code == 403:
+            msg += (
+                "\n\nIf this issue persists, generate a Github API token and store"
+                " it as the environment variable `GITHUB_TOKEN`:\n"
+                "https://github.blog/2013-05-16-personal-api-tokens/"
+            )
```
matnad, line 990: Avoid string concatenations with `+` (this can have unwanted side effects due to not being type safe). Instead use f-strings like `msg = f"{msg}\n\nIf..."`.
I am fully aware that this is not done as rigorous as it should be throughout the code base and you can feel free to change any instances you come accross ;-).

brownie/project/main.py
```diff
@@ -958,3 +961,36 @@ def _stream_download(download_url: str, target_path: str) -> None:
...
+        raise ConnectionError(msg)
+    elif r.json().get("default_branch") is None:
```
matnad, line 992: I would prefer `elif "default_branch" not in r.json():`

matnad (approved):
Thanks a lot for implementing the requested changes and thanks so much for contributing!
Code-wise this lgtm, but I do have some remarks on style, see below.

Maybe @iamdefinitelyahuman can give some input as well on the style policy and if this needs to be changed. Approval from me 💯

skellet0r:
@matnad thanks for the remarks, I definitely agree with them as they help with readability down the line, and are in line with the style of the project. I implemented the corrections via d860bb50.
Would be happy to hear @iamdefinitelyahuman's remarks as well, more input is always appreciated.

brownie/project/main.py
```diff
@@ -958,3 +959,37 @@ def _stream_download(download_url: str, target_path: str) -> None:
...
+    ---------
+    mix_name : str
+        Name of a brownie-mix repository without -mix appended.
+
+    Returns
+    -------
+    str
+        The default branch name on github.
+    """
+    REPO_GH_API = "https://api.github.com/repos/brownie-mix/{}-mix".format(mix_name)
```
iamdefinitelyahuman, line 977:
```suggestion
    REPO_GH_API = f"https://api.github.com/repos/brownie-mix/{mix_name}-mix"
```

brownie/project/main.py
```diff
@@ -958,3 +959,37 @@ def _stream_download(download_url: str, target_path: str) -> None:
...
+            msg_lines = (
+                msg,
+                "\n\nIf this issue persists, generate a Github API token and store",
+                " it as the environment variable `GITHUB_TOKEN`:\n",
+                "https://github.blog/2013-05-16-personal-api-tokens/",
+            )
+            msg = "".join(msg_lines)
+        raise ConnectionError(msg)
+    elif "default_branch" not in r.json():
+        msg = "API results did not include {}'s default branch".format(mix_name)
```
iamdefinitelyahuman, line 992:
```suggestion
        msg = f"API results did not include {mix_name}'s default branch"
```

iamdefinitelyahuman (commented):
Thank you for your contribution!  Overall I'm happy with the implementation, just a couple places where I'd echo @matnad 's comments re: preference for f-strings over `.format`.  Once those are modified I'm happy to see this merge :rocket:

matnad (approved):
LGTM
Thanks again for the contribution and the prompt changes. 
I will add a test for this once we have a new mix repo that doesn't use master.

iamdefinitelyahuman (approved):

Issue #746, opened 2020-09-08
Fork mode as a context manager


iamdefinitelyahuman:
### Overview
When using brownie in a live environment, it should be possible to temporarily enter into `--fork` via a context manager.

### Specification
Add a context manager `brownie.forked`:

```
with brownie.forked:
    scary_transactions()
```

* Upon entering the context, ganache launches and forks against whatever live environment brownie is using.  The active network shifts to fork mode, but all existing objects persist.
* When the context exits, it cleans up by removing any `TransactionReceipt` or `Contract` objects, and correcting nonces and balances.
* A kwarg can be included so that if no exception is raised during the context manager, the same transactions are then broadcasted on the mainnet.  Bonus points if they're sent using `required_confs=0`.

A lot of the logic for this should already exist, I think it's possible with some minor modifications to `chain.revert`, `chain.undo` and `chain.redo`.  The hardest part will be the idea of having an active network over top of another active network.  I'm not sure if this is a feasible change without also breaking things - this might have to wait until `v2.0.0`.  But it's worth exploring right away as the functionality would be very useful.

WyseNynja:
Anyone put any more thoughts into this? This would be really handy for testing transactions before sending them to mainnet.

WyseNynja:
I'm trying some things

PR #923, opened 2021-01-06
[wip] Fork mode as a context manager 

WyseNynja:
### What I did

Related issue: #746

### How I did it

Added a `fork` context manager and modified the rpc to support multiple children.

### How to verify it

Something like this:
```
    with fork(unlock=str(account)) as fork_settings:
        print("we forked!")
        print("balance:", account.balance())

        with fork(**fork_settings['cmd_settings']):
            print("we forked the fork!")
            print("balance:", account.balance())
```

Or this:
```
    with fork(unlock=str(account)):
        print("we forked!")
        print("balance:", account.balance())

    with fork(unlock=str(account2)):
        print("we forked again!")
        print("balance:", account.balance())
```

this style is currently broken:
```
    with fork(unlock=str(account)) as fork_settings:
        print("we forked!")
        print("balance:", account.balance())

    print("we are back on mainnet")
    print("balance:", account.balance())
```

### Checklist

- [ ] make sure any global state gets cleaned up

### DIFF
--- a/brownie/network/main.py
+++ b/brownie/network/main.py
@@ -47,7 +47,7 @@ def connect(network: str = None, launch_rpc: bool = True) -> None:
                     )
                 rpc.attach(host)
             else:
-                rpc.launch(active["cmd"], **active["cmd_settings"])
+                rpc.launch(active["cmd"], force=False, **active["cmd_settings"])
         else:
             Accounts()._reset()
         if CONFIG.network_type == "live" or CONFIG.settings["dev_deployment_artifacts"]:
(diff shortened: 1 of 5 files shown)


WyseNynja:
I could probably change it so that nesting the context managers uses snapshots instead of spinning up another ganache. That would probably be much faster. This was easy though

fubuloubu:
This is a great idea

WyseNynja:
After using this (and some variations of it) on some private repos. This pattern does not work great. There are some threading issues when things shut down. I have an idea that uses `disconnect` and `connect` and `snapshot` and preserves accounts that I'm going to try next.

Issue #943, opened 2021-01-30
Hitting Infura Rate Limit when running test suite locally


omarish:
### Environment information

* `brownie` Version: 1.13.0
* `ganache-cli` Version: x.x.x
* `solc` Version: x.x.x
* Python Version: 3.8.5
* OS: win/osx/linux

### What was wrong?

* I'm working on a pull request to try and get this working with `web3.py==5.12.3` but am running into some issues when running the test suite locally.
* Specifically, I'll run the test suite like so: `docker-compose exec sandbox bash -c 'python -m pytest tests/ -x' and pretty quickly start to have test failures due to Infura rate limits.
* The failing test passes when I run it individually, because it doesn't run into the same rate limit: `docker-compose exec sandbox bash -c 'python -m pytest tests/network/contract/test_contract.py::test_from_explorer -x'

### How can it be fixed?

I'm wondering if there are any quick ways around this. Perhaps there's a docker image we could use that mocks Infura endpoints? Open to other suggestions.

iamdefinitelyahuman:
Hrmm..  this is a new one to me.  Maybe infura has reduced the number of requests they allow in a given time period? Or the API key in the CI has found it's way into too many other repos.  Will have a look.

PR #944, opened 2021-01-30
Bumps Web3.py to >=5.11.1,<=5.14.0

omarish:
### What I did

* Updated web3.py to `>=5.11.1,<=5.14.0`
* This is a draft PR and haven't added any new tests yet.

Related issue:

* https://github.com/eth-brownie/brownie/issues/943

### Checklist


iamdefinitelyahuman:
I generally prefer to strictly pin dependencies.  In the past, `web3` has made breaking changes in patch versions.

That said, happy to move to a newer version.  Just haven't had the bandwidth to deal with changes in `5.12.0` that were causing tests to fail.

omarish:
Understood, that makes sense and I think strict pinning is probably the safe strategy, especially if `web3` has previously made breaking changes in patch versions. 

Looks like some of the test failures you mentioned are showing up here already. Will try and work through these when I can.

iamdefinitelyahuman:
This was handled in #1064

omarish:
Amazing. Nicely done, @skellet0r!

Issue #873, opened 2020-12-02
ValueError: invalid literal for int() with base 10: 'LID'`


benjyz:
### Environment information

* `brownie` Version: master
* `ganache-cli` Version: v6.12.1 
* `solc` Version: 0.6.12
* Python Version: 3.9
* OS: linux

### What was wrong?

code compiles fine before

```
 File "/usr/local/lib/python3.8/dist-packages/brownie/_cli/__main__.py", line 64, in main
    importlib.import_module(f"brownie._cli.{cmd}").main()
  File "/usr/local/lib/python3.8/dist-packages/brownie/_cli/console.py", line 47, in main
    active_project = project.load()
  File "/usr/local/lib/python3.8/dist-packages/brownie/project/main.py", line 702, in load
    return Project(name, project_path)
  File "/usr/local/lib/python3.8/dist-packages/brownie/project/main.py", line 164, in __init__
    self.load()
  File "/usr/local/lib/python3.8/dist-packages/brownie/project/main.py", line 217, in load
    self._compile(changed, self._compiler_config, False)
  File "/usr/local/lib/python3.8/dist-packages/brownie/project/main.py", line 90, in _compile
    build_json = compiler.compile_and_format(
  File "/usr/local/lib/python3.8/dist-packages/brownie/project/compiler/__init__.py", line 146, in compile_and_format
    build_json.update(generate_build_json(input_json, output_json, compiler_data, silent))
  File "/usr/local/lib/python3.8/dist-packages/brownie/project/compiler/__init__.py", line 309, in generate_build_json
    build_json[contract_name] = solidity._get_unique_build_json(
  File "/usr/local/lib/python3.8/dist-packages/brownie/project/compiler/solidity.py", line 246, in _get_unique_build_json
    pc_map, statement_map, branch_map = _generate_coverage_data(
  File "/usr/local/lib/python3.8/dist-packages/brownie/project/compiler/solidity.py", line 349, in _generate_coverage_data
    pc += int(pc_list[-1]["op"][4:])
ValueError: invalid literal for int() with base 10: 'LID'
```

### How can it be fixed?

no idea how to debug

iamdefinitelyahuman:
Can you give a minimal example so I can reproduce the error?

benjyz:
the failure came from this contract https://github.com/rapidtrade-io/launchswap/blob/master/contracts/PCert.sol

compiled fine on ropsten and remix, so wasn't sure where to start

[...]

PR #952, opened 2021-02-08
Fix non-push instructions

kirk-baird:
### What I did

There is an issue when processing instructions which are not `PUSH` that will lead to bad `int` conversion.

Related issue: #873 also related to PR #935 

The issue is almost identical to above but occurs in another location.

### Checklist


iamdefinitelyahuman (approved):
Good catch, ser!

Issue #371, opened 2020-03-12
Brownie console running in Vagrant  couldnt connect to ganache running on Host. 


GraphicalDot:
### Environment information

* `brownie` Version: 1.6.7
* `ganache-cli` Version: 2.1.2
* `solc` Version: 0.6.2
* Python Version: 3.7.6
* OS: linux

### What was wrong?
I am running ganache on my ubuntu host and running brownie project on a vagrant machine with ubuntu version18.04. 
Brownie couldn't connect to the ganache but web3.py can.
My config file for brownie is 
```
# Brownie configuration file
# https://eth-brownie.readthedocs.io/en/stable/config.html
network:
    default: development # the default network that brownie connects to
    settings:
        gas_limit: "auto"
        gas_price: "auto"
        persist: true
        reverting_tx_gas_limit: false  # if false, reverting tx's will raise without broadcasting
    networks:
        # any settings given here will replace the defaults
        development:
            host: http://172.28.128.1
            gas_price: 0
            persist: false
            reverting_tx_gas_limit: 6721975
            test_rpc:
                cmd: ganache-cli
                port: 3399
                gas_limit: 6721975
                accounts: 10
                evm_version: istanbul
                mnemonic: brownie
        # set your Infura API token to the environment variable WEB3_INFURA_PROJECT_ID
        mainnet:
            host: https://mainnet.infura.io/v3/$WEB3_INFURA_PROJECT_ID
        goerli:
            host: https://goerli.infura.io/v3/$WEB3_INFURA_PROJECT_ID
        kovan:
            host: https://kovan.infura.io/v3/$WEB3_INFURA_PROJECT_ID
        rinkeby:
            host: https://rinkeby.infura.io/v3/$WEB3_INFURA_PROJECT_ID
        ropsten:
            host: https://ropsten.infura.io/v3/$WEB3_INFURA_PROJECT_ID
        classic:
            host: https://www.ethercluster.com/etc
        kotti:
            host: https://www.ethercluster.com/kotti
pytest:
    # these settings replace the defaults when running pytest
    gas_limit: 6721975
    default_contract_owner: true
    reverting_tx_gas_limit: 6721975
    revert_traceback: true
compiler:
    evm_version: null
    minify_source: false
    solc:
        version: null
        optimize: true
        runs: 200

```
The port is open on host to accept new connections which can be checked with netstat. 
```
tcp        0      0 127.0.0.1:2222          0.0.0.0:*               LISTEN      -                   
[...]

PR #972, opened 2021-02-24
Support for dockerized RPC clients + Hostname resolution when attaching to RPC clients

mariuspod:
### What I did
I've extended the check that is determining if a RPC client is already running to also query the network connections in addition to  the process list. The error I've observed on a `Ubuntu 18.04 LTS` was that only localhost RPC clients could be detected with the existing code. Even dockerized RPC clients bound to the `127.0.0.1` could not be attached.

### How I did it
I've adapted the logic to first try the process list and then the network connections.
I've also implemented hostname resolution which becomes handy when the RPC client is bound to a specific hostname instead of IP which is often the case when used in docker-compose environments.

### How to verify it
```bash
docker run -p 8545:8545 trufflesuite/ganache-cli
```

Then in another terminal

```bash
brownie console
``` 

### Checklist


iamdefinitelyahuman:
I think this fixes #371

iamdefinitelyahuman (approved):
Very cool, thanks for your contribution

PR #996, opened 2021-03-18
Flag to run tests against project outside of current working directory

### LINKED ISSUE #994: Allow testing from another working directory
I want to use brownie pytest plugin, but our testing framework runs pytest from another working directory.
Currently, the pytest plugin uses check_for_project('.').

A possible solution is to add an environment variable with the real path of the project, and make check_for_project() search there as well.

iamdefinitelyahuman:
### What I did
Add a commandline flag `--brownie-project` to force loading a project outside of the current working directory when using the brownie pytest plugin.

Closes #994 

### How I did it
Within `test/plugin.py`, I've added the `_get_project_path` method to find the current project path.  Because the path required prior to `pytest_addoption`, it must be extracted directly from `sys.argv`.  It's a bit hacky but it solves the issue.

### How to verify it
Try it.

### DIFF
--- a/brownie/test/plugin.py
+++ b/brownie/test/plugin.py
@@ -2,6 +2,7 @@
 
 import sys
 from pathlib import Path
+from typing import Optional
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


brownie/test/plugin.py
```diff
@@ -12,9 +13,23 @@
 from brownie.utils import color
 
 
+def _get_project_path() -> Optional[Path]:
+    if "--brownie-project" not in sys.argv[:-1]:
```
spapinistarkware, line 17: Note: This means `--brownie-project=abcd` won't work, which is also a standard way of passing args. But It's probably good enough.

iamdefinitelyahuman: Ahh true..  that should be easy enough to support as well.  Good catch!

spapinistarkware: I am wondering, why not just add the flags in pytest_addoption, regardless of whether we are in a project dir or not?
I mean, if the user doesn't want to use brownie, he can just disable the plugin

Issue #850, opened 2020-11-20
dropped and replaced transaction hangs brownie while repeatedly hitting infura 18,091 times with eth_getTransactionReceipt


sabotagebeats:
### Environment information

* `brownie` Version: v1.11.0
* `ganache-cli` Version: v6.10.1 (ganache-core: 2.11.2)
* `solc` Version: 
* Python Version: 3.8.5 
* OS: win 10

### What was wrong?

Please include information like:

sent a transaction
transaction was dropped and replaced on etherscan
no error, just repeatedly hi infura 18,091 while waiting for transaction to confirm for 4 hours

### How can it be fixed?

Fill this in if you know how the bug could be fixed.
if transaction is dropped and replaced it should work with the replaced transaction

sabotagebeats:
![image](https://user-images.githubusercontent.com/27985126/99758172-d9fae100-2aa5-11eb-9966-4a5ab7c6d814.png)

iamdefinitelyahuman:
Oh my :grimacing: Was the replacement transaction broadcast within Brownie or via another program?

I _think_ this was fixed in #846 but I'll have a look again to be sure.

sabotagebeats:
I believe it's from using the same account key in different instances of brownie console simultaneously so the nonce is not getting updated.

sabotagebeats:
I will try updating brownie and also try using different keys for different simultaneously executing transactions

sabotagebeats:
Haven't experienced this since recent versions. Closing issue and will reopen if I encounter the problem again.

PR #997, opened 2021-03-18
Middlewares and Caching

iamdefinitelyahuman:
### What I did
* generalized logic for adding middlewares to the brownie `web3` object
* a basic implementation of for RPC-call caching

### How I did it
Middlewares are automagically loaded from all files within `brownie/network/middlewares`. The only requirement is that they subclass `BrownieMiddlewareABC`. They are automatically added upon connection and removed at disconnect.  The exact spec is fairly well documented within the natspec and comments.

The cache has two components:

* A short-term cache that lasts as long as the current block. This is useful because it means elsewhere in the code, there's no longer an onus to think about "am I hitting this endpoint too often?" - which gets exponentially more complex when you start using threading (such as with tx replacement). Related - #850
* A long-term cache that persists between sessions. Currently this is only storing `eth_getCode` under specific conditions (non-zero return value, no presence of the `SELFDESTRUCT` opcode). Looking at the Infura dashboard for my most commonly used API key, this endpoint accounts for >20% of the network traffic from Brownie, so it seems like a good first candidate. With the base logic in place for this cache it should be possible to expand it to other endpoints.

### DIFF
--- a/brownie/network/middlewares/__init__.py
+++ b/brownie/network/middlewares/__init__.py
@@ -1,0 +1,114 @@
+import functools
+import importlib
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import Callable, Dict, List, Optional
+
+from web3 import Web3
+
+
+class BrownieMiddlewareABC(ABC):
+
+    """
+    Base ABC for all middlewares.
+
+    This class must be inherited in order for a middleware to be discovered
+    and added to the `web3` object upon connecting to a network.
+    """
+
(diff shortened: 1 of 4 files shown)


drewwells:
How do I make use of this in a script? I'm pulling a lot of information from the network and would like to keep it locally in this cache.

iamdefinitelyahuman:
At present this isn't exposed in the public API, sorry.  You could look at the code in `network/middlewares` and find a way to hack it, but it might not be pretty.

Issue #1000, opened 2021-03-19
Add support for python-dotenv at the config level


PatrickAlphaC:
### Overview
I'd love to see brownie support a `.env` file for environment variables. For new engineers, or a project that doesn't want to keep a ton of environment variables running, a `.env` file is a warm welcome. Truffle and Hardhat (the javascript smart contract frameworks) have really nice config files that allow them to work with .env files. A bit of a quality of life improvement.

Otherwise, we have to call in the `.env` file in all our script manually, seems pretty redundant to do so. 

### Specification
Not sure how this is normal done in yaml, but would like to see something like:
`dotenv: true`

And then they can load via a `.env` file. We could just use https://pypi.org/project/python-dotenv/ to load at the start of all the scripts.

### Dependencies
https://pypi.org/project/python-dotenv/

What do we think?

iamdefinitelyahuman:
I see no harm in checking for the existence of this file when loading a project, and processing it if it's found.

omarish:
Would it make sense if the value of the `dotenv` key in the YAML config is a path instead of a boolean? That way you can specify if you're using a different dotenv.

PatrickAlphaC:
Ooooo. That's much better

omarish:
Cool. And @iamdefinitelyahuman, do you think the dotenv file should be a) auto-loaded, or b) loaded only if `dotenv:` is specified in the YAML config?

PatrickAlphaC:
I vote for loading only if it's specified.

skellet0r:
Loading only if explicitly opted in via the YAML config sounds better to me. Sort of how docker-compose allows one to load an env-file or multiple. Plus, security wise might be better to opt-in 🤔 

https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option

omarish:
I agree. I'd just like to get buy-in from @iamdefinitelyahuman before building something out. 

When it comes to crypto projects, I'm an advocate for configuration over convention. Writing an extra line in a config file is a great price to pay, especially if it means avoiding unexpected configuration behavior.

iamdefinitelyahuman:
Sounds good to me! I agree that the explicitness of adding a line to the config is better than potential accidentally... some config settings.

omarish:
Excellent. Another realization is that this should probably support interpolating envvars in the config, like this:

## `.env file`:

```
DEFAULT_NETWORK=development
```

## `brownie-config.yaml`:

```yaml
networks:
[...]

PR #1012, opened 2021-03-23
feat: brownie config supports env vars

omarish:
This feature adds POSIX-style environment variable expansion to Brownie's config. You can now access environment variables, like you would `os.environ`, or, define them using the new top-level `dotenv` flag brownie's configuration. 

# What I did

* Adds `dotenv:` key to `brownie-config.yml`
* Reads environment vars from both `os.environ` as well as a `.env` file (when specified)
* Supports default values, like so: `${THE_REAL_DEAL:-dentacoin}`, or `${NEEDS_MORE_COWBELL:-true}`
* Supports basic value inference for `int` and `bool`.

Related issue: #1000 

# Demos:

## System env vars:

```sh
export SHOW_COLORS=true
```

`brownie-config.yml`

```yaml
console:
  show_colows: ${SHOW_COLORS}
```

## Using a `.env` file

`.env`:

```ini
BUILD_DIR=/dev/teapot
```

`brownie-config.yml`

```yaml
dotenv: .env
project_structure:
  build: ${BUILD_DIR}
networks:
  default: ${DEFAULT_NETWORK:-development}
```

### How to verify it

See the tests.

### Checklist

- [x] Few more failing tests
- [x] Docs

### DIFF
--- a/brownie/_config.py
+++ b/brownie/_config.py
@@ -248,7 +248,7 @@ def _load_project_compiler_config(project_path: Optional[Path]) -> Dict:
     return compiler_data
 
 
-def _load_project_envvars(project_path):
+def _load_project_envvars(project_path: Path) -> Dict:
     config_vars = dict(os.environ)
     if CONFIG.settings.get("dotenv"):
         dotenv_path = CONFIG.settings["dotenv"]
--- a/brownie/_config.py
+++ b/brownie/_config.py
@@ -11,11 +11,13 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 import yaml
+from dotenv import dotenv_values
 from hypothesis import Phase
 from hypothesis import settings as hp_settings
 from hypothesis.database import DirectoryBasedExampleDatabase
 
 from brownie._singleton import _Singleton
+from brownie._expansion import expand_posix_vars
 
 __version__ = "1.14.3"
 
@@ -187,6 +189,17 @@ def _load_project_config(project_path: Path) -> None:
     """Loads configuration settings from a project's brownie-config.yaml"""
     config_path = project_path.joinpath("brownie-config")
     config_data = _load_config(config_path)
+    config_vars = _load_project_envvars(project_path)
+
+    if "dotenv" in config_data:
+        if not isinstance(config_data["dotenv"], str):
+            raise ValueError(f'Invalid value passed to dotenv: {config_data["dotenv"]}')
+        env_path = project_path.joinpath(config_data["dotenv"])
+        if not env_path.is_file():
+            raise ValueError(f"Dotenv specified in config but not found at path: {env_path}")
+        config_vars.update(dotenv_values(dotenv_path=env_path))  # type: ignore
+        config_data = expand_posix_vars(config_data, config_vars)
+
     if not config_data:
         return
 
@@ -217,6 +230,8 @@ def _load_project_config(project_path: Path) -> None:
 
     CONFIG.settings._unlock()
     _recursive_update(CONFIG.settings, config_data)
+    _recursive_update(CONFIG.settings, expand_posix_vars(CONFIG.settings, config_vars))
+
     CONFIG.settings._lock()
     if "hypothesis" in config_data:
         _modify_hypothesis_settings(config_data["hypothesis"], "brownie", "brownie-base")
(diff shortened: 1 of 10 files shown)


tests/project/test_brownie_config.py
```diff
@@ -128,3 +131,42 @@ def test_validate_cmd_settings():
...
+    test_brownie_config = f"""
+    dotenv: {env_filename}
+    networks:
+        default: development
+        development:
+            gas_limit: 6543210
+    """
+    env_path = testproject._path.joinpath(env_filename)
+    with env_path.open("w") as fp:
+        fp.write(test_dotenv_config)
```
omarish, line 155: Should probably use `TemporaryFile`, but I'd like to have a file that's a sibling to the config, since I think that's the most common way the `dotenv` directive will be used.

omarish:
Thanks! Will try my best to get this done as soon as I can.

omarish:
Ok, I think this is coming together now and nearing completion. I'm still seeing a lot of test failures but I have a feeling they're all from the same root cause. 

I added some nice touches in there, like recursive parsing, envier defaults, and basic type inference. Ended up spending more time on this than I had initially planned 😄. 

image

Will try and wrap this up tomorrow if I can block off a few more hours.

omarish:
Ok, going to dive into this now. Will tidy up and minimize the diff.

omarish:
Ok, I figured out what was going on.

* I was setting some nonsensical values to test env var propagation. Example: https://github.com/eth-brownie/brownie/pull/1012/commits/6ae54d675a453958289c982ceb868b40e9457bba
* I changed the tests to be more in-line with reasonable usage, and I'm back on the right path again and things seem to be behaving normally again.

Some takeaways:

* We might want to put in some timeouts if connecting to a default network takes more than a reasonable amount of time (like 5 minutes). I think that's why the previous builds were hanging and taking ~6h before Github Actions timed them out.

As for the most recent failures, I believe I'm just getting rate-limited (sorry about that). Either way, my optimistic engineering self thinks that this should work when I try it again a little bit later on. Going to mark it ready for review to start getting feedback; @iamdefinitelyahuman and maintainers: feel free to re-run tests when you take a look at this PR and I think everything will be passing. I'll also check again in a couple of hours.

omarish:
Also @PatrickAlphaC for sure re: example with RPC_URL. Where in the docs do you think it makes the most sense?

brownie/_expansion.py
```diff
@@ -0,0 +1,57 @@
+import re
+
+from dotenv.variables import parse_variables
+
+
+def expand_posix_vars(dct, variables):
```
iamdefinitelyahuman, line 6: Took me a moment to understand what `dct` and `variables` are here.  Some more expressive naming, mypy types, and/or an "arguments" section in the natspec might help.

brownie/_expansion.py
```diff
@@ -0,0 +1,57 @@
...
+    https://github.com/theskumar/python-dotenv/commit/17dba65244c1d4d10f591fe37c924bd2c6fd1cfc
+
+    We need this layer here so we can explicitly pass in variables;
+    python-dotenv assumes you want to use os.environ.
+    """
+
+    if value is None:
+        return None
+    if not isinstance(value, (str)):
+        return value
```
iamdefinitelyahuman, line 36: The first check here seems redundant, I believe the 2nd, broader check would always catch the first condition.

Also the brackets around `str` have no effect without a trailing comma.

brownie/_config.py
```diff
@@ -233,6 +248,19 @@ def _load_project_compiler_config(project_path: Optional[Path]) -> Dict:
     return compiler_data
 
 
+def _load_project_envvars(project_path):
```
iamdefinitelyahuman, line 251: Add a MyPy type here

iamdefinitelyahuman (commented):
Generally this looks really good!  Just a few minor nitpicks / questions.

iamdefinitelyahuman (approved):

PatrickAlphaC:
AMAZING

omarish:
Hey @kkom: good question. It's been a little bit since I did this PR so I don't 100% remember if there was a reason I had pinned it that way. I think this one is @iamdefinitelyahuman's call, but here's what I think:

1. I read python-dotenv's changelog between 0.16.x and the latest and I also re-read the code I wrote for this PR. Fortunately I don't think I'm calling any private APIs/doing anything that would require the restriction to be that strict.
2. If we had infinite time I'd suggest testing it on `0.{17,18,19,20}.x`, but I'd 👍 your PR to remove the `0.17.0` constraint entirely if it just worked on `0.20` as well and deal with potential issues for 0.17-0.19 in the future if they arise.

kkom:
Sounds good! Btw, what's your thought on the upper constraint? Technically according to semver anything can happen before you hit `1.0.0` so minor releases don't need to be backwards compatible AFAIK. But again, if I replaced `<0.17.0` with `<0.21.0` we'd be at square one in a few minor releases. So pragmatically I'm inclined to perhaps do `<1.0.0`, even though it's not 100% proper according to a pure interpretation of semantic versioning. WDYT?

omarish:
Great question. Yeah there's the essence of semver and there's how it's
actually done. I'd be comfortable with `<0.21.0` if it was only my say, but
I'm going to yield to @iamdefinitelyahuman on what we actually end up doing.

Omar Bohsali
http://omarish.com | @omarish <http://twitter.com/omarish>
(650) 394-6627


On Sat, May 21, 2022 at 4:27 AM, Konrad Komorowski ***@***.***

kkom:
Sure thing! My 2 cents are the following – before finalizing the PR I'd take a look at the usages of the library, if they look very basic and unlikely to change – I'd recommend to do `<1.0.0`. Worst case some forward compatibility patch will need to be issues for brownie, but I honestly suspect that relaxing the dependency graph will overall be a nice benefit for most users.

Issue #918, opened 2021-01-02
IndexError raised when passing an element of brownie.accounts as the exclude argument to brownie.test.strategy 


skellet0r:
### Environment information

* `brownie` Version: 1.12.3, 1.12.2, 1.12.1
* `ganache-cli` Version: 6.12.1
* `solc` Version: 0.8.0+commit.c7dfd78e.Linux.g++
* Python Version: 3.8.3, 3.9.1
* OS: linux

### What was wrong?

Command: `brownie test`
Code that caused failure:
```python
from brownie import accounts
from brownie.test import given, strategy

@given(
    amount=strategy("uint256", max_value="1000 ether"),
    to=strategy("address", exclude=accounts[0]),
)
def test_transfer_success_returns_true(token, accounts, amount, to):
    assert token.transfer(to, amount, {"from": accounts[0]}) == True
```
Error output:
```bash
$ brownie test tests/test_transfer.py
Brownie v1.12.3 - Python development framework for Ethereum

=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.8.3, pytest-6.0.1, py-1.10.0, pluggy-0.13.1
rootdir: /home/skelletor/projects/CloudToken
plugins: eth-brownie-1.12.3, hypothesis-5.41.3, forked-1.3.0, web3-5.11.1, xdist-1.34.0
collected 0 items / 1 error                                                                                                                                                                                       

===================================================================================================== ERRORS ======================================================================================================
_____________________________________________________________________________________ ERROR collecting tests/test_transfer.py _____________________________________________________________________________________
tests/test_transfer.py:8: in <module>
    to=strategy("address", exclude=accounts[0]),
E   IndexError: list index out of range
============================================================================================= short test summary info =============================================================================================
FAILED tests/test_transfer.py - IndexError: list index out of range
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================================================================ 1 error in 0.15s =================================================================================================
```

### Extra Information

[...]

PR #1013, opened 2021-03-24
address hypothesis strategy exclude fix - issue 918

sabotagebeats:
### What I did
the `hypothesis` strategies will now accept `address` `exclude` which previously didn't work

Related issue: # 
[918](https://github.com/eth-brownie/brownie/issues/918)

### How I did it
1. connect to the network
2. with a `for` loop, load each accounts from `kwargs['exclude']` into a list `exclude` and then use that list to clobber `kwargs['exclude']`
3. disconnect from network

### How to verify it
the `address exclude` will now accept a list of account indexes 
e.g. `exclude=[0,1,2]` for `accounts[0]`, `accounts[1]`, and `accounts[2]`

```
#!/usr/bin/python3
from brownie import network, accounts
from brownie.test import given, strategy
from brownie import network

@given(
receiver=strategy('address', exclude=[0,9]),
amount=strategy('uint256', max_value=10**18),
)
def test_transfer_adjusts_receiver_balance(accounts, token, receiver, amount):
    balance = token.balanceOf(receiver)
    token.transfer(receiver, amount, {'from': accounts[0]})
    assert receiver != accounts[0], 'receiver is accounts[0]'
    assert receiver != accounts[9], 'receiver is accounts[9]'

    assert token.balanceOf(receiver) == balance + amount
```

### Checklist

- [❌] I have confirmed that my PR passes all linting checks
- [✅] I have included test cases
- [✅] I have updated the documentation
- [❌] I have added an entry to the changelog


sabotagebeats:
`ERROR: InvocationError for command ~/brownie/.tox/lint/bin/isort --check-only --diff --recursive brownie tests --skip brownie/__init__.py (exited with code 1)`

sabotagebeats:
`CHANGELOG.md` updated and all checks pass with `tox`

brownie/test/strategies.py
```diff
@@ -158,6 +159,15 @@ def strategy(type_str: str, **kwargs: Any) -> SearchStrategy:
     if type_str == "fixed168x10":
         return _decimal_strategy(**kwargs)
     if type_str == "address":
+        if "exclude" in kwargs:
+            if not network.is_connected():
```
iamdefinitelyahuman, line 163: This will trigger a connection/disconnection for every test that uses `exclude`, which adds about 3 seconds per test on my machine.  It could also have issues if ganache isn't using deterministic addresses for the local accounts.

I think it would be better to modify the [deferred strategy](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.deferred) logic on line 82 and apply the exclusion when the lambda function executes to create the data set. This approach wouldn't add the overhead of connecting/disconnecting.

iamdefinitelyahuman (changes requested):

brownie/test/strategies.py
sabotagebeats: I had originally tried to put there but I wasn't able to get it working. I think that's because I was using it wrong. I'll give it another try and see if I can get it working there.

sabotagebeats: @iamdefinitelyahuman did I complete these changes satisfactorily or is there more I need to do?

CHANGELOG.md
```diff
@@ -6,11 +6,16 @@ This changelog format is based on [Keep a Changelog](https://keepachangelog.com/
...
+<<<<<<< HEAD
+### Fixed
+- Support `hypothesis` strategies now accept `address` `exclude` ([#1013](https://github.com/eth-brownie/brownie/pull/1013))
+=======
 - Add polygon network integration ([#1119](https://github.com/eth-brownie/brownie/pull/1119))
 - Fixed subcalls to empty accounts not appearing in the subcalls property of TransactionReceipts ([#1106](https://github.com/eth-brownie/brownie/pull/1106))
 
 ### Added
 - Added `LocalAccount.sign_message` method to sign `EIP712Message` objects ([#1097](https://github.com/eth-brownie/brownie/pull/1097))
+>>>>>>> 4a399b1323b75226a801c16af1cc170273506bef
```
fubuloubu, line 18: Rebase commands still in file

brownie/test/strategies.py
```diff
@@ -76,9 +76,25 @@ def _decimal_strategy(
 
 
 @_exclude_filter
-def _address_strategy(length: Optional[int] = None) -> SearchStrategy:
+def _address_strategy(length: Optional[int] = None, exclude_by_index=None) -> SearchStrategy:
```
fubuloubu, line 79:
```suggestion
def _address_strategy(length: Optional[int] = None, exclude_by_index: Optional[List[int]]=None) -> SearchStrategy:
```

docs/tests-hypothesis-property.rst
```diff
@@ -81,7 +81,7 @@ It is possible to supply multiple strategies via ``@given``. In the following ex
     from brownie.test import given, strategy
 
     @given(
-        to=strategy('address', exclude=accounts[0]),
+        to=strategy('address', exclude=[0]),
```
fubuloubu, line 84:
```suggestion
        to=strategy('address', exclude_by_index=[0]),
```

fubuloubu (commented):
Please run `git rebase -i master` to squash commits together for review

CHANGELOG.md
fubuloubu: https://codeinthehole.com/guides/resolving-conflicts-during-a-git-rebase/

PR #1030, opened 2021-04-01
fix: balance contract member override

banteg:
### What I did

Fixed a regression introduced [here](https://github.com/eth-brownie/brownie/commit/ed7adacd008d2a633c47c701ccf905954fe66a40#diff-9d15ba572e3097f4a9461dc8fe0a0dfc6d7f6015e935930526ebfb4ba0fa0cefL781).

Related issue: #

### How I did it

Reverted an erroneous refactor to its previous state.

Also added renaming `balance` Contract class member to `wei_balance` on name clash instead of rendering it inaccessible.

### How to verify it

`Contract('0x5dbcF33D8c2E976c6b560249878e6F1491Bca25c').balance()` on mainnet should be non-zero.

### Checklist

### DIFF
--- a/brownie/network/contract.py
+++ b/brownie/network/contract.py
@@ -778,9 +778,10 @@ class _DeployedContractBase(_ContractBase):
         if name == "balance":
             warnings.warn(
                 f"'{self._name}' defines a 'balance' function, "
-                f"'{self._name}.balance' is unavailable",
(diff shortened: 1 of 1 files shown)


poolpitako:
Thanks for the fix :+1:
Tested with bsc and it works.

iamdefinitelyahuman (approved):

Issue #1011, opened 2021-03-21
Multicall context manager


iamdefinitelyahuman:
### Overview
Brownie should offer first-class support for [multicall](https://github.com/makerdao/multicall) via a context manager:

```
with multicall:
    a = foo.bar()
    b = foo.barfoo()
    c = bar.foobar(a)
    d = bar.otherbar()
```

### Specification
While the context manager is open, a middleware intercepts all `eth_call` requests and returns objects which represent the result of the call once it completes.  When the context manager closes, or there is an attempt to use one of the pending call objects, a single multicall is made to get the actual values.

In the above example `a` and `b` would end up joined as a single multicall that executes when `a` is needed for `c`.  Then `c` and `d` would happen as a multicall when the context manager exits.

We can bundle the [`MultiCall2`](https://github.com/yearn/yearn-exporter/blob/c59a8207cb8721cb4264cbae741a9438b3486cba/contracts/Multicall2.sol) implementation as used by yearn, which uses try/catch to prevent calls from failing.  Attempting to interact with object for the specific result that failed in the multicall should raise a `ValueError`. (Or possibly the object simply returns `None`? Maybe this behavior is determined via a kwarg in the context manager?)

It is important to consider thread safety during the implementation.  The context manager must introspect to be aware of which thread it is running on, and not also batch `eth_call` operations coming from unrelated threads.

To ensure consistent behaviour across network, we can use Geth's [state override](https://geth.ethereum.org/docs/rpc/ns-eth#override-example) feature to simulate the existence of a multicall contract for the purposes of the call.  This is badass in so many ways I'm not even sure where to begin :smiling_imp: For dev networks where this isn't possible, deploy multicall silently the first time it's required.

banteg:
This is an amazing idea! A few caveats:
- ganache-cli doesn't expect a third argument 
- as does web3py https://github.com/ethereum/web3.py/issues/1921

omarish:
I've been thinking about this over the past couple days and wanted to float an idea. Would we definitely want it to automatically resolve calls (i.e know to evaluate `a` and `b` before getting `c`)?

One possible design alternative, just to be exhaustive:

```python
# (A) This works
with multicall:
    a = foo.bar()
    b = foo.barfoo()

# (B) This raises an error
with multicall:
    a = foo.bar()
    b = foo.barfoo()
    c = bar.foobar(a)

# (C) This is how we'd want to write (B)
with multicall:
    a = foo.bar()
[...]

PR #1033, opened 2021-04-04
[wip] feat: multicall context manager

omarish:
### What I did

This is a rough first draft at a multicall context manager. It doesn't work yet but sharing it here to get feedback before I go deeper into implementation.

Related issue: #1011

### What Remains

- [ ] Thread safety
- [ ] Making it so that using the `MulticallContextManager` (`multicall`) enables the middleware

### Checklist

### DIFF
--- a/brownie/data/contracts/Multicall2.sol
+++ b/brownie/data/contracts/Multicall2.sol
@@ -1,0 +1,72 @@
+pragma solidity >=0.5.0;
+pragma experimental ABIEncoderV2;
+
+/// @title Multicall2 - Aggregate results from multiple read-only function calls
+/// @author Michael Elliot <mike@makerdao.com>
+/// @author Joshua Levine <joshua@makerdao.com>
+/// @author Nick Johnson <arachnid@notdot.net>
+
+contract Multicall2 {
+    struct Call {
+        address target;
+        bytes callData;
+    }
+    struct Result {
+        bool success;
+        bytes returnData;
+    }
+
+    function aggregate(Call[] memory calls) public returns (uint256 blockNumber, bytes[] memory returnData) {
+        blockNumber = block.number;
+        returnData = new bytes[](calls.length);
+        for(uint256 i = 0; i < calls.length; i++) {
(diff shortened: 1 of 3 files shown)


banteg:
I've played with it quite a lot too, a few thoughts:

Multicall as middleware won't work, it requires patching the web3 internals too deeply, and possibly replacing web3 provider.

A better way might be yielding a caller decorator from the context manager which intercepts function calls, generates the calldata using `encode_input` and storing a `decode_output` reference. This could potentially also work with automatic flushing of the queue as was originally envisioned since you have the idea of the argument types before yielding futures.

What I haven't solved so far is keeping the API the same. I don't like that it adds this `.result()` requirement.

This is what I got so far:
```python
from concurrent.futures import Future
from contextlib import contextmanager
from functools import partial


class Caller:
    def __init__(self, queue, contract=None, func=None):
        self.queue = queue
        self.contract = contract
        self.func = func

    def __getattr__(self, name):
        return Caller(self.queue, self.contract, getattr(self.contract, name))

    def __call__(self, *args, **kwds):
        future = Future()
        future.call = [str(self.contract), self.func.encode_input(*args, **kwds)]
        future.decode_output = self.func.decode_output
        self.queue.append(future)
        return future


@contextmanager
def multicall_context():
    """
    with multicall_context() as caller:
        response = caller(contract).func(args)

    response.result()
    """
    queue = []
    yield partial(Caller, queue)  # caller(contract).func(args).result()

    block, results = multicall.aggregate.call([request.call for request in queue])

    for request, result in zip(queue, results):
        request.set_result(request.decode_output(result))
```

A full example:
```python
def main():
    vault = Contract("0x5dbcF33D8c2E976c6b560249878e6F1491Bca25c")

    with multicall_context() as caller:
        assets = caller(vault).balance()  # can't call result before exiting
        pps = caller(vault).getPricePerFullShare()

    print(assets.result().to('ether') * pps.result().to('ether'))
```

spinoch:
> What I haven't solved so far is keeping the API the same. I don't like that it adds this .result() requirement.

Could use `ContextManager.__exit__` to take care of that (?)

PR #1043, opened 2021-04-06
Support for Hardhat Network

iamdefinitelyahuman:
### What I did
Add preliminary support for [Hardhat Network](https://hardhat.org/hardhat-network/) as a testing environment.

To test this, you must currently install the `hardhat@dtt` package. The required features for this integration will not merge to the main hardhat repo until after the Berlin fork, in the mean time this PR should also be considered a WIP.

To test, create a Brownie project and then install hardhat inside it:

```bash
brownie init
npm install --save-dev hardhat@dtt
touch hardhat.config.js
```

You can now use hardhat as a dev environment with `--network hardhat` or hardhat fork mode using `--network hardhat-fork`.

### How I did it
Most of the groundwork for this was laid in previous PRs (#997, #998). The integration is made possible via a new backend module and a hardhat-specific middleware.

### How to verify it
Try it! And please report any issues you encounter.

### TODO
Before this merges, the test suite should be modified to target both hardhat and ganache. We also need to start use this in our daily lives and smoke out any weird edge cases or other oddities.

Closes #1024
Closes #1025
Closes #1026
Closes #1027

PR #1044, opened 2021-04-07
Hardhat compatibility fixes

banteg:
### What I did

Screen Shot 2021-04-07 at 07 47 01

### Remaining issues
- brownie.exceptions.RPCRequestError: the tx doesn't have the correct nonce. account has nonce of: 0 tx has nonce of: 1
- Parallel tests don't work (`-n auto`)


iamdefinitelyahuman:
Merged most of the work here into #1043

PR #1062, opened 2021-04-16
more BSCSCAN_TOKEN

WyseNynja:
### What I did

I found a spot where we were still using ETHERSCAN_TOKEN instead of BSCSCAN_TOKEN.

### How to verify it

```
$ export BSCSCAN_TOKEN=xxx
$ brownie --network bsc-main console
>>> Contract.from_explorer("0x2170ed0880ac9a755fd29b2688956bd959f933f8")
```

### Checklist


iamdefinitelyahuman (approved):
Looking at the diff history, I don't think this is a regression - it was just never added. Guessing bscscan recently modified something in their API due to an increase in requests?

poolpitako:
yep, that's my bad. I am using `ETHERSCAN_TOKEN` for the bsc token.

Issue #1048, opened 2021-04-13
Update web3 to be compatible with web3-flashbots


WyseNynja:
### Overview

I am wanting to use https://github.com/flashbots/web3-flashbots.git with a brownie project.

However, I get this error when trying to install both:

```
The conflict is caused by:
    eth-brownie 1.14.4 depends on web3==5.11.1
    flashbots 0.3.0 depends on web3<6.0.0 and >=5.17.0
```

### Specification

Can we upgrade the web3 in brownie to >=5.17.0?

Any known blockers on that? I half remember web3 upgrades being difficult.

WyseNynja:
For now, I think I might be able to do `export BROWNIE_LIB=1`

WyseNynja:
`export BROWNIE_LIB=1` is working for me

WyseNynja:
Actually that upgrades too much and breaks `brownie test`. I'm trying now with running `pip-compile --upgrade-package web3` for brownie.

iamdefinitelyahuman:
Fixed by #1064

Issue #1010, opened 2021-03-21
Dependencies for hypothesis, pytest and web3 are outdated


iamdefinitelyahuman:
### Overview
The latest versions of `hypothesis`, `pytest` and `web3` all break functionality somewhere in Brownie.

### Specification
Someone needs to bump each dependency to the latest version, fix the failing tests, and play around in the console / run some large test suites to make sure there is no further weirdness.  This could be a big-ish job, given the level of hackiness and touching-private-APIs that brownie does with `pytest` and `hypothesis` :grimacing:

Maxence-L:
Does that mean testing doesn't work with v1.14.5 ?

PR #1064, opened 2021-04-19
Upgrade web3 dependency

skellet0r:
### What I did

Upgrade web3 dependency to 5.18.0, replace soon to be deprecated methods, and add additional middleware to maintain backwards compatibility (SolidityError/ContractLogicError)

A majority of this is just some grepppppping to change method names from camelCase to snake_case :) 

Related issue: #1048 #1010 

Additional Info:

In the 5.13.0 release of web3, there is a new error formatter middleware for `eth_call` and `eth_estimateGas` rpc methods, this formatter instead of passing along the rpc response, instead it raises a ContractLogicError with a (pretty meh) message, but granted it's for end users. For brownie, all we have to really do is bypass that middlware by adding our own prior to it, and raise a ValueError with the appropriate RPCError dict. This is then used to raise a VirtualMachineError when a transaction reverts etc ....

References: ethereum/web3.py#941 ethereum/web3.py#1814

### How I did it

Used pip-compile -P web3 {,requirements-dev.in,requirements-windows.in}

### How to verify it

Since this is supposed to be mainly upgrading the dependencies and not adding any new features, the entire test suite should pass across all builds.

I've also tested using multiple different repos which utilize brownie (the curvefi org of tests)

### Checklist

- [x] I have confirmed that my PR passes all linting checks and tests


iamdefinitelyahuman (approved):
Wow, did not expect such a huge diff.

Thank you for doing this :pray:

skellet0r:
Closes #1048

PR #1072, opened 2021-04-30
read network name from script directly

### LINKED ISSUE #1071: specify network within the script
Right now, a script runs with the extra parameter `--network` on the command line, I'm seeing that if we could have script specify its own network configuration, it would be better. So I don't need to remember which script matches with which network.

I suggest we add another function name within script:

```
def setup_network():
    pass
```

It returns a network configuration that can be used for connection.

guanqun:
fix #1071

I'm not sure how to write unit tests for this, if you could give any suggestion, that'll be great.

### What I did

Related issue: #1071 

### Checklist

### DIFF
--- a/brownie/network/account.py
+++ b/brownie/network/account.py
@@ -460,6 +460,44 @@ class _PrivateKeyAccount(PublicKeyAccount):
                 "If you wish to broadcast, include `allow_revert:True` as a transaction parameter.",
             ) from None
 
+    def deploy_raw(self, contract_name: str, hex_str: str):
+        with self._lock:
+            try:
+                gas_price = 0
+                gas_limit = Wei(12000000)
+
+                txid = self._transact(  # type: ignore
+                    {
+                        "from": self.address,
+                        "value": Wei(0),
+                        "nonce": self._pending_nonce(),
+                        "gasPrice": gas_price,
+                        "gas": gas_limit,
+                        "data": HexBytes(hex_str),
+                    },
+                    False,
+                )
+                exc, revert_data = None, None
+            except ValueError as e:
+                exc = VirtualMachineError(e)
+                if not hasattr(exc, "txid"):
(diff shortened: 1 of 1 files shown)


guanqun:
When I add a new entry in changelog, under which section should I put?

brownie/_cli/run.py
```diff
@@ -40,11 +40,12 @@ def main():
         active_project.load_config()
         print(f"{active_project._name} is the active project.")
 
-    network.connect(CONFIG.argv["network"])
-
     path, _ = _get_path(args["<filename>"])
     path_str = path.absolute().as_posix()
 
+    network_from_script = _get_network_from_script(path)
```
iamdefinitelyahuman, line 46: One potential issue here - within `_get_network_from_script` you call to `_import_path`, which means the module gets imported prior to connecting to the network.  If anything happens in the global namespace that requires an active network connection, the script will now break.

The first solultion that comes to mind for me: instead of importing the module, parse the ast and then search for the related `Assign` node.  Something like this (tho for sure it'll need more validation):

```python
with path.open() as fp:
    module_ast = ast.parse(fp.read())
    network_name = next(
        (i.value.value for i in module_ast.body if isinstance(i, ast.Assign) and i.targets[0].id == "brownie_default_network"),
        None
    )
    if network_name is not None: ...
```

iamdefinitelyahuman (commented):
Sorry for very slow reply, was on vacation...

brownie/_cli/run.py
guanqun: Thank you for the review. I agree with your concern.

Parsing the ast could only help us search for the specific pattern of the assignment. Once it needs evaluation of the expression, it's hard to do in pure AST.

guanqun:
Sorry I don't have enough time on this. Closing this PR.

PR #1079, opened 2021-05-07
feat: allow users to overwrite 'host' and 'explorer' in network settings

guanqun:
### What I did
I'd like to have a setting for mainnet. Infura is not the only option here. I can specify "host" with another eth RPC provider.
similar to etherscan, I'd like to have another url, as the original one is blocked here.

### How I did it
I add the code in _config.py

### Checklist

### DIFF
--- a/brownie/_config.py
+++ b/brownie/_config.py
@@ -94,6 +94,19 @@ class ConfigContainer:
 
             network["cmd_settings"]["fork"] = os.path.expandvars(network["cmd_settings"]["fork"])
 
+        if (
+            key == "live"
+            and id_ in self.settings["networks"]
+        ):
+            overwrite_settings = self.settings["networks"][id_]
+
+            if "host" in overwrite_settings:
+                # overwrite the host settings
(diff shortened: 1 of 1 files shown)


skellet0r:
@guanqun I'm sure the network docs need to shed some more light on this, but you can actually overwrite network settings already.
There are a few routes one can take:
1. Via the command line, you can run `brownie networks modify`
2. Adding a network-config.yaml in your repository, and then importing via `brownie networks import` (you'll have to set the overwrite flag to `True`)
3. Lastly, going directly to the `~/.brownie/network-config.yaml` file and manually correcting the entries of interest.

The simplest option (and most future proof) is option 2. Just remember when committing your network-config, not to include and secrets.

[Example of a network-config.yaml](https://github.com/eth-brownie/brownie/blob/master/brownie/data/network-config.yaml)

> tl;dr: run `brownie networks -h` and you'll see a bunch of available options for managing your available networks

guanqun:
Thank you @skellet0r , I wasn't aware of this usage. I'll check this out.

PR #1106, opened 2021-06-04
Fix: Subcalls to empty accounts

### LINKED ISSUE #1105: Bug: Calls to empty accounts do not appear in tx.subcalls
### Environment information

* `brownie` Version: 1.14.6
* `ganache-cli` Version: 6.12.2
* `solc` Version: 0.8.4
* Python Version: 3.8.9
* OS: linux

### What was wrong?

Low level calls to addresses which have no deployed contract do not show up in `tx.subcalls`.

Below is some example contracts in both solidity and vyper, as well as a sample set of test cases. 

*contracts/SolDummy.sol*
```solidity
// SPDX-License-Identifier: MIT
pragma solidity >=0.6.0;

contract SolDummy {
    address constant TARGET = 0xD0660cD418a64a1d44E9214ad8e459324D8157f1;

    function foo_call() external {
        TARGET.call(abi.encodeWithSignature("bar()"));
    }

    function foo_static_call() external {
        TARGET.staticcall(abi.encodeWithSignature("bar()"));
    }

    function foo_delegate_call() external {
        TARGET.delegatecall(abi.encodeWithSignature("bar()"));
    }
}
```

*contracts/VyDummy.vy*
```python
# @version 0.2.12


TARGET: constant(address) = 0xD0660cD418a64a1d44E9214ad8e459324D8157f1


@external
def foo_call():
    raw_call(TARGET, method_id("bar()"))


@external
def foo_static_call():
    raw_call(TARGET, method_id("bar()"), is_static_call=True)


@external
def foo_delegate_call():
    raw_call(TARGET, method_id("bar()"), is_delegate_call=True)
```

*tests/test_call.py*
```python
import pytest


@pytest.fixture(scope="module")
def alice(accounts):
    return accounts[0]


@pytest.fixture(scope="module")
def sol_dummy(alice, SolDummy):
    return SolDummy.deploy({"from": alice})


@pytest.fixture(scope="module")
def vy_dummy(alice, VyDummy):
    return VyDummy.deploy({"from": alice})


@pytest.fixture(scope="module", params=(0, 1), ids=("SolDummy", "VyDummy"))
def dummy(request, sol_dummy, vy_dummy):
    return [sol_dummy, vy_dummy][request.param]


def test_call(alice, dummy):
    tx = dummy.foo_call({"from": alice})
    assert len(tx.subcalls) == 1


def test_static_call(alice, dummy):
    tx = dummy.foo_static_call.transact({"from": alice})
    assert len(tx.subcalls) == 1


def test_delegate_call(alice, dummy):
    tx = dummy.foo_delegate_call({"from": alice})
    assert len(tx.subcalls) == 1
```
When the above tests are ran, they all fail, when in fact they should pass as there is exactly 1 low level sub call in each of the dummy functions.

### How can it be fixed?

This needs to be fixed in the `brownie/network/transaction.py` file, specifically in [this area](https://github.com/eth-brownie/brownie/blob/master/brownie/network/transaction.py#L780).

skellet0r:
### What I did

Minor patch of the logic in transactions.py to handle direct calls/staticcalls/delegatecalls to empty accounts. 

Fixes #1105 

### How I did it

Added to the if statement that determines whether a call is being made to a contract, and handled calls to the datacopy address from vyper.

### How to verify it

There is a parameterized test verifying functionality in both solidity and vyper contracts.

### Checklist

### DIFF
--- a/brownie/network/transaction.py
+++ b/brownie/network/transaction.py
@@ -781,6 +781,7 @@ class TransactionReceipt:
         last_map = {0: _get_last_map(self.receiver, self.input[:10])}  # type: ignore
         coverage_eval: Dict = {last_map[0]["name"]: {}}
 
+        datacopy_precompile = "0x0000000000000000000000000000000000000004"
         call_opcodes = ("CALL", "STATICCALL", "DELEGATECALL")
         for i in range(len(trace)):
             # if depth has increased, tx has called into a different contract
@@ -823,6 +824,10 @@ class TransactionReceipt:
                         self._subcalls[-1]["calldata"] = calldata.hex()
                 elif calldata:
                     self._subcalls[-1]["calldata"] = calldata.hex()
+
+                if str(self._subcalls[-1]["from"]) == datacopy_precompile:
+                    caller = self._subcalls.pop(-2)["from"]
+                    self._subcalls[-1]["from"] = caller
 
             # update trace from last_map
             last = last_map[trace[i]["depth"]]
--- a/brownie/network/transaction.py
+++ b/brownie/network/transaction.py
@@ -271,7 +271,8 @@ class TransactionReceipt:
     def subcalls(self) -> Optional[List]:
         if self._subcalls is None:
             self._expand_trace()
-        return self._subcalls
+        subcalls = filter(lambda s: not _is_call_to_data_copy(s), self._subcalls)  # type: ignore
+        return list(subcalls)
 
     @trace_property
     def trace(self) -> Optional[List]:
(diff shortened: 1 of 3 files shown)


brownie/network/transaction.py
```diff
@@ -775,9 +776,13 @@ def _expand_trace(self) -> None:
         last_map = {0: _get_last_map(self.receiver, self.input[:10])}  # type: ignore
         coverage_eval: Dict = {last_map[0]["name"]: {}}
 
+        datacopy_precompile = "0x0000000000000000000000000000000000000004"
```
iamdefinitelyahuman, line 779: Why do you single out this specific precompile and not the others?

skellet0r: Weak sauce on my part, changed now to use `re` and match against all the current pre-compiles as per geth's codebase

iamdefinitelyahuman (approved):

PR #1140, opened 2021-07-06
Bind GUI report dropdowns to currently displayed contract

benber86:
Currently, the values of the report dropdowns in the brownie GUI are loaded once on startup and remain static afterwards. The values can be changed by the user, but neither the values nor the highlights are updated when another contract is selected - even if that contract has no corresponding report.

This can become inconvenient in different ways: 
- If someone is just starting to write tests or only running a specific subset of tests, when checking for coverage of branches or statements through the GUI, if no highlights are available for the selected contract it is not clear whether that is due to the absence of a report or to the tests failing to cover any part of the contract for that report type. 
- When switching between contracts, the report and report type dropdowns retain their previous values but the highlights themselves are cleared. One needs to reselect a report type (even if it is the same) to get the proper highlights for the current contract. This can likewise create confusion as it can easily appear that a specific report type has been selected and that tests have failed to cover any of the code in the contract.
- More of a cosmetic issue, but upon startup, one can select a report and highlight options without having selected a contract.

### What I did

I tied the report and report type values to the active tab of the tkinter notebook where the contract code is displayed. If no reports are available for the currently displayed contract, it is made explicit: the report dropdown displays "(No Report)" and becomes disabled. When one switches between tabs or contract, the dropdown values are updated depending on the availability of a report for said contract. When switching between two contracts that both have reports available, the user's selection of report type is retained but the highlights in the notebook are updated to those of the newly selected contract.

### How I did it

Made some modification to the dropdown's code to check that the currently displayed contract has a corresponding report. Linked the values to the active tab in the notebook in `source.py` rather than to the `get_active_contract()` method from the root.

### How to verify it

Run only a subset of a project's test suite so as not to cover all available contracts. Open brownie gui and navigate between the different contracts.

### Known issues

No tests (there are currently no tests for the gui)

### Checklist

### DIFF
--- a/brownie/_gui/report.py
+++ b/brownie/_gui/report.py
@@ -1,19 +1,41 @@
 #!/usr/bin/python3
 
 import warnings
+from pathlib import Path
 
 from .bases import SelectBox
 
 
 class ReportSelect(SelectBox):
     def __init__(self, parent):
-        super().__init__(parent, "(No Reports)", [])
+        super().__init__(parent, "", [])
+
+    def show(self):
+        self.grid()
+
+    def hide(self):
+        self.grid_remove()
+
+    def set_values(self, contract):
         reports = self._root().reports
-        if not reports:
+        values = []
+        for report in sorted(reports):
+            if contract in reports[report][report]:
+                values.append(report)
+
+        if not values:
+            self.set("(No Reports)")
             self.config(state="disabled")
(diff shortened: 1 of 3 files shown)


iamdefinitelyahuman:
Very cool!

iamdefinitelyahuman (approved):

PR #1179, opened 2021-08-03
feat: eip1559 transactions support

banteg:
### What I did

Add support for specifying EIP-1559 style transactions.

### How I did it

`transact` method not accepts `max_fee` and `priority_fee` arguments which are mutually exclusive with legacy `gas_price` argument. I went for shorter names instead of `max_fee_per_gas` and `max_priority_fee_per_gas` to keep Brownie users' sanity.

It also adds `chain.base_fee` to look up the base fee of the latest block. It it used when a user only specifies `priority_fee`, in which case `max_fee` is inferred as `base_fee * 2 + priority_fee`, which should be enough to get the transaction through within 6 blocks of raising base fee (actually 1.125⁶ is needed, explore if that would be a better constant here).

### Checklist

### DIFF
--- a/brownie/network/account.py
+++ b/brownie/network/account.py
@@ -689,7 +689,7 @@ class _PrivateKeyAccount(PublicKeyAccount):
             silent = bool(CONFIG.mode == "test" or CONFIG.argv["silent"])
 
         try:
-            if gas_price is not None:
+            if max_fee is None and priority_fee is None:
                 gas_price, gas_strategy, gas_iter = self._gas_price(gas_price)
             else:
                 gas_strategy, gas_iter = None, None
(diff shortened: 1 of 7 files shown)


brownie/network/account.py
```diff
@@ -944,26 +944,24 @@ def _apply_fee_to_tx(
...
+        if max_fee or priority_fee:
+            raise ValueError("gas_price and (max_fee, priority_fee) are mutually exclusive")
+        tx["gasPrice"] = web3.toHex(gas_price)
+        return tx
 
-    elif max_fee is not None and priority_fee is not None:
-        if priority_fee > max_fee:
-            raise InvalidTransaction("priority_fee must not exceed max_fee")
+    if priority_fee is None:
+        raise InvalidTransaction("priority_fee must be defined")
```
banteg, line 954: Check if these invalid transaction checks can be ommited and caught downstream in web3.py

brownie/network/account.py
```diff
@@ -944,26 +944,24 @@ def _apply_fee_to_tx(
 ):
     tx = tx.copy()
 
-    if gas_price is not None and (max_fee or priority_fee):
-        raise ValueError("gas_price and (max_fee, priority_fee) are mutually exclusive")
+    if gas_price is not None:
+        if max_fee or priority_fee:
+            raise ValueError("gas_price and (max_fee, priority_fee) are mutually exclusive")
+        tx["gasPrice"] = web3.toHex(gas_price)
```
banteg, line 950: Check if toHex is really required here, I think web3.py does them automatically now.

brownie/network/account.py
```diff
@@ -944,26 +944,24 @@ def _apply_fee_to_tx(
...
-        if priority_fee > max_fee:
-            raise InvalidTransaction("priority_fee must not exceed max_fee")
+    if priority_fee is None:
+        raise InvalidTransaction("priority_fee must be defined")
 
     # no max_fee specified, infer from base_fee
-    elif max_fee is None and priority_fee is not None:
+    if max_fee is None:
         base_fee = Chain().base_fee
         max_fee = base_fee * 2 + priority_fee
```
banteg, line 960: 2 gwei is a constant [from here](https://notes.ethereum.org/@barnabe/rk5ue1WF_), but in reality we want to predict an optimal priority fee. In the future we can use [`eth_feeHistory` api](https://github.com/zsfelfoldi/feehistory) for that, but it's not implemented in many clients at the moment.

banteg (commented):
Few things to address:
- We should prefer newer transaction type if the chain supports it.
- This can be checked by checking if the latest a block has `baseFeePerGas`.
- In this case none of (gas_price, max_fee, priority_fee) defined should result in priority_fee = TBD and max_fee = 2 * base_fee + priority fee.
- You now need to fetch a receipt to get `effectiveGasPrice`.
- If the chain doesn't support EIP-1559, fall back to gas price and gas strategies.
- In the future expose useful info from `eth_feeHistory` and use it for gas estimation.
- Document the change in `eth_call` behavior. Specifying non-zero `gas_price` below `base_fee` can result in unexpected breakage. Hardhat also has some known quirks which might prevent proper testing contracts which use `BASEFEE` opcode.

iamdefinitelyahuman:
This PR gets us as far as "it works" - it's far from perfect and there are likely unresolved issues, but it's far enough along that we can start using the new transaction types and get a feel for what's good and what's missing

iamdefinitelyahuman (approved):

brownie/network/account.py
iamdefinitelyahuman: Make an issue and come back when the clients are ready, I guess.

banteg:
Need to wait for the next web3.py release before coining a Brownie release, it's broken currently.
https://github.com/ethereum/web3.py/pull/2099

Issue #1201, opened 2021-08-13
`with reverts()` not working as expected on hardhat network


morelazers:
### Environment information

* `brownie` Version: 1.16.1
* `solc` Version: 0.8.7
* Python Version: 3.8.3
* OS: osx

### What was wrong?

Running
```
brownie test --network hardhat
```
is causing some tests to fail. Removing `--network hardhat` results in them passing.

Have created a reproduction here: https://github.com/morelazers/brownie-hardhat-bug-repro

```solidity
pragma solidity 0.8.7;

contract Impl {
	function shouldThrow (uint256 greaterThanTen) public pure {
		require(greaterThanTen > 10, "Number not greater than 10");
	}
}
```

```py
from brownie import Impl, accounts, reverts

ERR_MSG = "Number not greater than 10"

def test_shouldThrow():
	impl = Impl.deploy({ 'from': accounts[0] })
	with reverts(ERR_MSG):
		impl.shouldThrow(1)


def test_shouldThrowWithoutWithStatement():
	impl = Impl.deploy({ 'from': accounts[0] })
	impl.shouldThrow(1)
```

PR #1204, opened 2021-08-16
Support hardhat revert strings

morelazers:
Solves #1201 for me, I'm not really a python dev so not sure if this is how to solve the issue "properly"

`tox -e lint` fails because of something in `brownie/convert/datatypes.py`


iamdefinitelyahuman (commented):
This addresses the issue, but it does so fairly far away from the root of the problem.  I've opened another PR, #1218, to apply the fix within a middleware immediately after receiving the data from Hardhat.

PR #1209, opened 2021-08-17
pre and post hook

benjyz:
### What I did

allow for pre and post hooks, so that the entire build process looks like this

1) pre compile hook. setup any files and needed utilities etc.
2) compile hook. run through each file and process it
3) post hook. final steps after compilation e.g. generate a flat file

this also allows to e.g. template contract code and other processing. an idea for the future would be a hooks.yaml which maps the custom config to the steps.

- [x ] I have confirmed that my PR passes all linting checks
- [-] I have included test cases
- [-] I have updated the documentation
- [-] I have added an entry to the changelog

### DIFF
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -953,6 +953,9 @@ def _load_sources(project_path: Path, subfolder: str, allow_json: bool) -> Dict:
     if project_path.joinpath("brownie_hooks.py").exists():
         hooks = importlib.import_module("brownie_hooks")
 
+    if hasattr(hooks, "pre_hook"):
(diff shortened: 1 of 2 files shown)


skellet0r:
Gotta fix the lint, also … might be good to finally add some docs so people know about these hooks. Think you could add those in too ?

benjyz:
sure, for lint I run black. I thought had run it last time, let me see if it works now

docs - made a suggestion, not sure its right path

benjyz:
for the lint I run black. not sure what broke

docs - added a suggestion

PR #1232, opened 2021-09-02
fix: add lock when flushing the pending call queue Multicall

skellet0r:
### What I did

Add a lock when flushing the multicall pending call queue, to resolve concurrency bug

### How I did it

Added a lock around the part that swaps the bytecodes

### How to verify it

Run this script

```python
from brownie import multicall, ZERO_ADDRESS, Contract
from concurrent.futures import ThreadPoolExecutor


factory = Contract('0xB9fC157394Af804a3578134A6585C0dc9cc990d4')


def get_pools():
    with multicall:
        pools = [factory.pool_list(i) for i in range(int(factory.pool_count()))]
    return pools


def get_balances(pool):
    with multicall:
        coins = factory.get_coins(pool)
        balances = factory.get_balances(pool)
        decimals = factory.get_decimals(pool)

    return {
        coin: balance / 10 ** dec
        for coin, balance, dec in zip(coins, balances, decimals)
        if coin != ZERO_ADDRESS
    }


def main():
    pools = get_pools()
    datas = list(ThreadPoolExecutor().map(get_balances, pools))
    print(dict(zip(pools, datas)))
```

at this commit c2c288ace26c4e0987883017c637ec36216e290d <- fix
and at this commit b82bc03b04b92e5f9ecc0ed2bc391d43c94f4dc8
You'll notice the difference in execution time

Thanks to @banteg for spotting

### DIFF
--- a/brownie/network/multicall.py
+++ b/brownie/network/multicall.py
@@ -1,7 +1,7 @@
 import json
 from collections import defaultdict
 from dataclasses import dataclass
-from threading import get_ident
+from threading import Lock, get_ident
 from types import FunctionType, TracebackType
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -43,6 +43,8 @@ class LazyResult(Proxy):
 class Multicall:
     """Context manager for batching multiple calls to constant contract functions."""
 
+    _lock = Lock()
+
     def __init__(self) -> None:
         self.address = None
         self.block_number = None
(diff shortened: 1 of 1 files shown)


brownie/network/multicall.py
```diff
@@ -66,13 +68,14 @@ def _flush(self, future_result: Result = None) -> Any:
...
             # or this result has already been retrieved
             return future_result
-        ContractCall.__call__.__code__ = getattr(ContractCall, "__original_call_code")
-        results = self._contract.tryAggregate(  # type: ignore
-            False,
-            [_call.calldata for _call in self._pending_calls],
-            block_identifier=self.block_number,
-        )
-        ContractCall.__call__.__code__ = getattr(ContractCall, "__proxy_call_code")
+        with self.lock:
```
iamdefinitelyahuman, line 71: I'd place this lock at line 67 instead..  this way if one thread is blocked while another thread clears `_pending_calls`, the 2nd thread then doesn't try to do anything once it gets the lock.

brownie/network/multicall.py
```diff
@@ -43,6 +43,8 @@ def __repr__(self) -> str:
 class Multicall:
     """Context manager for batching multiple calls to constant contract functions."""
 
+    lock = Lock()
```
iamdefinitelyahuman, line 46:
```suggestion
    _lock = Lock()
```

:)

iamdefinitelyahuman (approved):
lfg

PR #1233, opened 2021-09-02
fix: change lock acquire/release in multicall

skellet0r:
### What I did

Ended up testing 4 some different ways to handle this (some combinations as well).

- Lock enter/exit of context
- Lock pending calls access / bytecode swap
- Add dictionary with per-thread data (pending calls, block number)
- Add a separate thread for flushing

Easiest and quickest was just locking on enter/exit (we'd need to use a RLock to handle reentrancy within the same thread though, and then also handle more reentrancy issues (like maintaining a queue of block numbers). However, opting for locking on queue/bytecode access + per-thread dictionary was what I ended up going with. For now it gets the job done with the least amount of headache.

Will definitely need to revisit this along with reentrancy in the future.

### DIFF
--- a/brownie/network/multicall.py
+++ b/brownie/network/multicall.py
@@ -57,7 +57,7 @@ class Multicall:
         ContractCall.__call__.__code__ = self._proxy_call.__code__
 
     @property
-    def block_number(self):
+    def block_number(self) -> int:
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman (approved):
gm

PR #1254, opened 2021-09-14
Auto generated project name includes numerics

Scowley4:
This change would allow numerics to be included in project names.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 __pycache__
+*.swp
 .vscode
 .history
 .coverage
--- a/brownie/exceptions.py
+++ b/brownie/exceptions.py
@@ -168,6 +168,7 @@ class ProjectAlreadyLoaded(Exception):
 class ProjectNotFound(Exception):
     pass
 
+
 class BadProjectName(Exception):
     pass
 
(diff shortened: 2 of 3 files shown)


Scowley4:
I guess now that I'm looking a little more, is the `isalpha` meant to prevent projects that _start_ with a numeric? 

Python imports can by any [identifier](https://docs.python.org/2/reference/lexical_analysis.html#identifiers):

`(letter|"_") (letter | digit | "_")*`

If the changed line is protecting for python purposes, maybe throw an error if the project name actually starts with a numeric. I'm just thinking some projects use versioning in their smart contracts. For example: [UniswapV2](https://github.com/Uniswap/uniswap-v2-core) vs [UniswapV3](https://github.com/Uniswap/uniswap-v3-core).

iamdefinitelyahuman:
Yeah, the check could definitely be modified so that `isalpha` is only applied to the first digit.

Scowley4:
Sorry, didn't realize renaming branch would mess up the PR :/

I've added an exception and written a check for the first character of a project/package name, throwing `BadProjectName` if the character isn't alphabetic.

iamdefinitelyahuman (approved):

Issue #1224, opened 2021-08-31
Wei rounding does not appear to match solidity


jeffywu:
### Environment information

* `brownie` Version: 1.16.0
* Python Version: 3.8.5
* OS: linux

### What was wrong?

```
>>> x
2200094942857964307
>>> y
1000000
>>> z
100000000
>>> Wei(Wei(x) * Wei(y) / Wei(z))
22000949428579644
>>> x
2200094942857964307
```

The error appears to be in the last digit, I was expecting x to be rounded down to:
22000949428579643

PR #1260, opened 2021-09-16
Add floor division to Wei

Scowley4:
### What I did:
I added floor division to the Wei datatype in order to better match Solidity.

Related issue: #1224 

### What's the problem
Brownie `Wei` class rounds differently from solidity because it uses true division (as inherited from python3.x `int`). Solidity uses floor division.

### Example (From #1224)
```python
# python
from brownie.convert import Wei

x = 2200094942857964307
y = 1000000
z = 100000000
out = Wei(Wei(x) * Wei(y) / Wei(z)) 
# 22000949428579644

# Essentially the same as:
out = Wei(int(x * y / z))
                  # ^ True division
```

```solidity
// solidity
contract MyContract{
    function division_testing() external view returns (uint out) {
        uint x = 2200094942857964307;
        uint y = 1000000;
        uint z = 100000000;
        out = x * y / z;
        // 22000949428579643
    }
}
```
----

### Thoughts

After playing with this for a bit and writing this up, I had some thoughts:

#### Overriding `__mul__` in this way causes all `Wei` calculations to be integers.

As I understand it, there aren't decimals in solidity, so while the below is "true", it may not be intuitive. 
```python
# (New code)
Wei(10) * 0.1 == 0
```
One thought would be throwing a `TypeError: can't multiply Wei by decimal`, but then `Wei(1) * 0.1` would fail and `0.1 * Wei(1)` would succeed, which is also confusing.

----

#### Use `//` operator instead
Could just opt to use the `//` (floor division or `__floordiv__`) instead. This would solve the original rounding problem above and none of the changes in this PR would be required.

----

Not sure where we want to go with this, and I'm very new to Brownie, so I don't have a deep understanding of where/how `Wei` is used.

### Checklist

### DIFF
--- a/brownie/convert/datatypes.py
+++ b/brownie/convert/datatypes.py
@@ -74,6 +74,15 @@ class Wei(int):
     def __sub__(self, other: Any) -> "Wei":
         return Wei(super().__sub__(_to_wei(other)))
 
+    def __floordiv__(self, other: Any) -> "Wei":
+        return Wei(super().__floordiv__(_to_wei(other)))
+
+    def __truediv__(self, other: Any) -> "Wei":
+        return self // _to_wei(other)
+
+    def __mul__(self, other: Any) -> "Wei":
+        return Wei(super().__mul__(_to_wei(other)))
+
     def to(self, unit: str) -> "Fixed":
         """
         Returns a converted denomination of the stored wei value.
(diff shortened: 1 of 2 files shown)


Scowley4:
The original problem is maybe not a "rounding" problem exactly, but more of a floating point problem. The exact result of true division would be `22000949428579643.07` but the closest floating points are `...9640, ...9644, ...9648`.

Issue #1138, opened 2021-07-05
Ignore commented out pragma lines


bernatfp:
### Overview

Brownie doesn't seem to ignore commented out `pragma` lines when it tries to find the best compiler to use. This is problematic for source files like this:

```
// pragma solidity >=0.7.5;
pragma solidity ^0.6.11;
```

In this example, Brownie will compile the interface with `solc 0.7.5` or greater, and fail with a message like:

```
interfaces/uniswap-v3/INonfungiblePositionManager.sol:6:1: ParserError: Source file requires different compiler version (current compiler is 0.7.6+commit.7338295f.Darwin.appleclang) - note that nightly builds are considered to be strictly less than the released version
pragma solidity ^0.6.11;
^----------------------^
```

### Specification

I think this can be solved by improving the robustness of the regular expression in https://github.com/eth-brownie/brownie/blob/master/brownie/project/sources.py#L214

PR #1320, opened 2021-10-30
#1138 fix: stop reading solidity version within comments

BobTheBuidler:
### What I did
Updated the relevant regex expression to ignore comments

Related issue: #1138 

### How I did it
Updated the relevant regex expression to ignore comments

### How to verify it
image

### Checklist

- [x] I have updated the documentation (doesn't seem relevant for this minor fix, didn't do this but did check box)

### DIFF
--- a/brownie/project/sources.py
+++ b/brownie/project/sources.py
@@ -211,7 +211,7 @@ def get_pragma_spec(source: str, path: Optional[str] = None) -> NpmSpec:
     Returns: NpmSpec object
     """
 
-    pragma_match = next(re.finditer(r"pragma +solidity([^;]*);", source), None)
+    pragma_match = next(re.finditer(r"(?:\n|^)pragma +solidity([^;]*);", source), None)
     if pragma_match is not None:
         pragma_string = pragma_match.groups()[0]
         pragma_string = " ".join(pragma_string.split())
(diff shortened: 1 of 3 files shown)


brownie/project/sources.py
```diff
@@ -211,9 +211,13 @@ def get_pragma_spec(source: str, path: Optional[str] = None) -> NpmSpec:
     Returns: NpmSpec object
     """
 
-    pragma_match = next(re.finditer(r"pragma +solidity([^;]*);", source), None)
+    pragma_match = next(re.finditer(r"(?<!([\*/ \t]))pragma +solidity([^;]*);", source), None)
```
banteg, line 214: A better fix would be to match on the start of line, i.e. `^pragma...`

banteg (changes requested):

brownie/project/sources.py
BobTheBuidler: I just tested this revision locally and using this method, I get 3 failing tests:

FAILED tests/project/compiler/test_solidity.py::test_build_json_keys - brownie.exceptions.PragmaError: No version pragma in 'path.sol'
FAILED tests/project/compiler/test_solidity.py::test_size_limit - brownie.exceptions.PragmaError: No version pragma in 'foo.sol'
FAILED tests/project/compiler/test_solidity.py::test_find_best_solc_version - brownie.exceptions.PragmaError: No version pragma in '<stdin>'

BobTheBuidler: since that didn't work, I tried some other varieties and found that this logic used for vyper also works for sol:
(?:\n|^)pragma +solidity([^;]*);

I've updated my branch accordingly

Issue #1190, opened 2021-08-09
brownie pm install fails if the version name contains an @


sapi-dlpf:
### Environment information

* `brownie` Version: 1.15.2
* OS: linux

### What was wrong?

Please include information like:

I'm trying to install Superfluid packages using brownie pm install, but the versions I need contain an "@" (like "ethereum-contracts@v1.0.0-rc.5"). So when i try 
brownie pm install superfluid-finance/ethereum-contracts@ethereum-contracts@v1.0.0-rc.5
Brownie fails to parse this string and gives me an error:
Invalid package ID. Must be given as [ORG]/[REPO]@[VERSION]

### How can it be fixed?

brownie should parse the string correctly, allowing the use of an @ inside the version identifier (or let the users escape this second @ somehow).

freddie71010:
Trying to bring some more visibility to this as there's a fix awaiting approval - https://github.com/eth-brownie/brownie/pull/1344

PR #1344, opened 2021-11-14
Allow dependency [VERSION] tagged with ("@")

veloboot:
Some GitHub projects use the at-sign ("@") in release version tags.  For dependencies and remappings, only split the [ORG]/[REPO]@[VERSION] on the first occurrence of the at-sign ("@") since [VERSION] may also contain the at-sign ("@").

### What I did

Merely use 1 as the 2nd parameter to the split function.

### DIFF
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -817,7 +817,7 @@ def _maybe_retrieve_github_auth() -> Dict[str, str]:
 
 def _install_from_github(package_id: str) -> str:
     try:
-        path, version = package_id.split("@")
+        path, version = package_id.split("@", 1)
         org, repo = path.split("/")
     except ValueError:
         raise ValueError(


altugbakan (approved):
Can we please add this? It fixes issue #1190 and would be helpful for importing some repositories like Superfluid.

freddie71010 (approved):

joshim2021 (approved):

freddie71010:
can we get this merged?

dinhbx-smartosc (approved):

dinhbx-smartosc (approved):
it fixed issue when package id contain '@' character.

dinhbx-smartosc (changes requested):
It fixed issue when package id contains multiple `@` character.

PR #1353, opened 2021-11-23
feat: calculate `CREATE` and `CREATE2` addresses from a deployed contract

skellet0r:
### What I did

Given that contracts can also be created by other contracts (via `CREATE` and `CREATE2`), I've added two new methods which allow the off-chain calculation of deployment addresses.

`get_deployment_address` - is an exact clone of the same method on account objects, and queries the address of a contract created with the `CREATE` opcode

`calculate_deterministic_address` - as the name suggests, calculates the address of a deterministically deployed contract via the `CREATE2` opcode.

### How to verify it

There are two new test files verifying the results of both new functions.

### Checklist


iamdefinitelyahuman (commented):
This seems very useful, but I worry about collisions with contract function names.  `nonce` in particular seems like something somebody will surely have used as a public var in a contract.

Could we add similar logic to how `balances` works where a collision results in a warning but doesn't bork anything?

WyseNynja:
We should also document that these "deterministic" addresses sometimes depend on things outside of the contract. For example, I've gotten different bytecode on my mac than on my linux machine. There's several things that seem to cause differences.

PR #1355, opened 2021-11-25
Use websockets >10.1 instead of pip compiled 9.1

pygeek:
websockets 9.1 is incompatible with python 3.10 due to the following error:
"TypeError: As of 3.10, the *loop* parameter was removed from Lock() since it is no longer necessary"

This change does not conflict with web3.py, which is the only required dependency that uses websockets, as it allows for websockets>=9.1.

### What I did

Add websocket>10.1 dependency in order to override default pip compile installation of websockets 9.1, which is the lowest compatible version allowed by web3.

Related issue: #

### How to verify it

To reproduce issue:
1. install etc-brownie using python 3.10
2. generate an account
3. Add and run script, which loads the account.

### Checklist


iamdefinitelyahuman (approved):

iamdefinitelyahuman:
Hrmm, gonna have to revert this unfortunately.  `web3py` pins to `websockets>=9.1,<10` so until they update we cannot.

PR #1365, opened 2021-12-05
brownie pm install: support version==latest

ydm:
`brownie pm install` (from git) now supports version=latest, e.g.
`brownie pm install OpenZeppelin/openzeppelin-contracts@latest`

### What I did

The list of available github tags is fetched anyways, so what I did is:
1. check if `version==latest`
2. sort tags by their semver order (also works for versions prefixed with `v`, i.e. `v1.0.0`)
3. pick the greatest

LMK if I need to (1) adapt the feature for ethpm packages, (2) update the documentation, (3) change log.

### DIFF
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -815,6 +815,26 @@ def _maybe_retrieve_github_auth() -> Dict[str, str]:
     return {}
 
 
+def _latest_tag(tags: List[str]) -> str:
+    if not tags:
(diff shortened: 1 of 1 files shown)


ydm:
Actually that's not a good patch as it makes simple commands like `brownie compile` perform GitHub requests on every run (since the install_path check takes place *after* fetching tags)...

Issue #1368, opened 2021-12-08
Make it possible to call "brownie.project.load" multiple times with the same project path.


zomglings:
### Overview
Currently, `project.load` raises a `ProjectAlreadyLoaded` error when you try to load a project from a path which already has a loaded project.

This makes it inconvenient to use directly in Python code.

It would be preferable if the following code ran without errors:
```
from brownie import project
project_path = "<path>"
project_1 = project.load(project_path)
project_2 = project.load(project_path)
assert project_1 is project_2
```

I am not sure if there are use cases where it may make sense for a user to force a reload (rebuild) of a project. If there are such cases, we could add a `force` keyword argument to make the following code run without errors:
```
from brownie import project
project_path = "<path>"
project_1 = project.load(project_path)
project_2 = project.load(project_path, force=True)
assert project_1 is not project_2
```

### Dependencies
Include links to any open issues that must be resolved before this feature can be implemented.

zomglings:
Suggested implementation: #1369

PR #1369, opened 2021-12-08
Improved project load

zomglings:
### What I did

Related issue: #1368 

### How I did it

Introduces the `raise_if_loaded` boolean argument, which is `True` by default. Default behaviour is the previous behaviour of `brownie`, but if `raise_if_loaded=False` them the loaders do not raise a `ProjectAlreadyLoaded` error when a caller tries to load a previously loaded project.

`project.load` quietly returns the previously loaded `Project` instance, and `project.main.Project.load` short-circuits (returning `None`).

### How to verify it
```bash
pytest tests/project/main/test_main_project.py::test_load_does_not_raise_when_raise_if_loaded_is_false
```

### Checklist

### DIFF
--- a/brownie/project/main.py
+++ b/brownie/project/main.py
@@ -747,15 +747,11 @@ def load(
             raise BadProjectName("Project must start with an alphabetic character")
         name = "".join(i for i in name.title() if i.isalnum())
 
-    already_loaded_project: Optional[Project] = None
     for loaded_project in _loaded_projects:
         if loaded_project._name == name:
-            already_loaded_project = loaded_project
-
-    if already_loaded_project is not None:
-        if raise_if_loaded:
-            raise ProjectAlreadyLoaded("There is already a project loaded with this name")
-        return already_loaded_project
+            if raise_if_loaded:
+                raise ProjectAlreadyLoaded("There is already a project loaded with this name")
+            return loaded_project
 
     # paths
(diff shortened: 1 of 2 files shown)


brownie/project/main.py
```diff
@@ -740,8 +746,16 @@ def load(project_path: Union[Path, str, None] = None, name: Optional[str] = None
...
+
+    already_loaded_project: Optional[Project] = None
+    for loaded_project in _loaded_projects:
+        if loaded_project._name == name:
+            already_loaded_project = loaded_project
+
+    if already_loaded_project is not None:
+        if raise_if_loaded:
+            raise ProjectAlreadyLoaded("There is already a project loaded with this name")
+        return already_loaded_project
```
jummy123, line 758: I _think_ you can shorten this a bit by nesting the `if` statement on 755 inside the first loop.

```
for loaded_project in _loaded_projects:
    if loaded_project._name == name:
        if raise_if_loaded:
            raise ProjectAlreadyLoaded("There is already a project loaded with this name")
        else:
            return loaded_project
```
Note: I haven't actually checked out the patch and ran the tests though.

zomglings: @jummy123 : You are correct, my friend. Fixed.

iamdefinitelyahuman (approved):

PR #1383, opened 2021-12-22
feat: adding a use_latest_patch option to solc compiler config

mariuspod:
### What I did

I've noticed that there's an issue with the chainlink contract `0x514910771AF9Ca656af840dff83E8264EcF986CA` wrt. to overloaded events. This contract is both `ERC-20` and `ERC-677` based and so it emits both types of events which have a different signature and can't be matched properly with log filters. AFAIS, overloaded events have been fixed in solidity compiler version `0.4.17`, the contract defines `0.4.16` in its pragma version. The fact that brownie doesn't compile solidity contracts with version `<0.4.22` causes it to use the given abi from the explorer etherscan which is also incomplete in terms of overloaded events. To overcome this issue and generate a more complete abi I've added a solc compiler setting that enables the download of the latest patch version of a given compiler based on the contract's version pragma.

### How I did it
In this change, I've implemented a solidity compiler config flag called `use_latest_patch`.
It can be turned on for all contracts by setting the value `true`:
```
compiler:
    solc:
        use_latest_patch: true
```

Alternatively, specific contracts can be added as a string list:
```
compiler:
    solc:
        use_latest_patch:
            - '0x514910771AF9Ca656af840dff83E8264EcF986CA'
```

In both cases the latest semver patch version of the solc compiler is downloaded and if it's >= 0.4.22 this causes local re-compilation of the contract with the latest compiler instead of the one which is hard-coded in the contract sources. 

For the chainlink contract, this fixes the issue that the abi contained only a single Transfer event definition and now includes both event definitions as the re-compilation is done locally.

### How to verify it
The contract 0x514910771AF9Ca656af840dff83E8264EcF986CA specifies the compiler version `0.4.16` in its source code.
The latest patch semver version for solidity is therefore `0.4.26`

If the flag is set to either true or if the contract is contained in the list, the `0.4.26` version will be used, otherwise the `0.4.16` will be used.

```
# with use_latest_patch = true
>>> Contract.get_solc_version("v0.4.16", "0x514910771AF9Ca656af840dff83E8264EcF986CA")
Version('0.4.26')

# with use_latest_patch = false
>>> Contract.get_solc_version("v0.4.16", "0x514910771AF9Ca656af840dff83E8264EcF986CA")
Version('0.4.16')
```
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/eth-brownie/brownie)
 ### Changed
 - Force files to be opened as UTF-8
+- Added a new solidity compiler setting `use_latest_patch` in brownie-config.yaml to use the latest patch version of a compiler based on the pragma version of the contract.
 
 ## [1.17.2](https://github.com/eth-brownie/brownie/tree/v1.17.2) - 2021-12-04
 ### Changed
--- a/brownie/network/contract.py
+++ b/brownie/network/contract.py
@@ -1091,11 +1091,13 @@ class Contract(_DeployedContractBase):
                 version = Version(compiler_str.lstrip("v")).truncate()
                 compiler_config = _load_project_compiler_config(Path(os.getcwd()))
                 solc_config = compiler_config["solc"]
-                if "use_latest_patch" in solc_config and solc_config["use_latest_patch"] is True:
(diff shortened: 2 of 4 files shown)


mariuspod:
Is there anything I could add to improve this PR ?

I'm using this functionality for capturing transfer events for the given chainlink contract. Without it, the decoding of the received events fail because they don't match the transfer topic from the cached abi.

iamdefinitelyahuman (dismissed):

PR #1411, opened 2022-01-26
Fix incremental compilation

danhper:
### What I did

Fix issue with the current incremental compilation causing all the contracts to recompile every time.

### How I did it

Improved the compiler comparison to avoid comparing things such as "v0.8.10" and "v0.8.10+commit.fc410830" or so.

### How to verify it

Compiling any project where the Solidity compiler version is specified should be enough to reproduce.

### Checklist


iamdefinitelyahuman (dismissed):

iamdefinitelyahuman (approved):

caenrigen:
@danhper @iamdefinitelyahuman Thanks a lot! Great speed up for big projects 😄

Issue #1092, opened 2021-05-27
Make Interfaces Ephemeral


banteg:
### What was wrong?

Brownie caches interfaces the same way it caches contract metadata from compiling from source. This can mess with the user expectations when they simply meant to poke at a contract to see if it responds to methods of a certain interface.

In Solidity interfaces are ephemeral, we should follow the same behavior and not associate a potentially incomplete or erroneous ABI with a contract address. Current workaround after using an incorrect interface is to call `Contract.from_explorer` on that address to overwrite the metadata.

Ephemeral interfaces would allow things like testing whether an unknown code at address follows a certain interface without worrying it would break things down the line. An example could be discerning between Uniswap/Balancer/Curve pools.

### How can it be fixed?

Don't cache `interface.Something(address)` as canonical ABI for the `address`, handle it like Solidity does.

iamdefinitelyahuman:
Yeah, this absolutely makes sense and shouldn't be that hard to implement.

banteg:
I looked a bit more into it. It seems removing this line is enough https://github.com/eth-brownie/brownie/blob/4a399b1323b75226a801c16af1cc170273506bef/brownie/network/contract.py#L1006

Are there any negative side effects to be aware of?

PR #1432, opened 2022-02-09
Make contract deployments optional, disabled for interfaces

mariuspod:
### What I did
I've disabled the persistence of interfaces into the deployments db.

Related issue: #1092

### How I did it
- Added a new function param `persist: bool = True` to several contract methods
- Passing `persist=False` for interfaces
- Added helper function `Contract.remove_deployment()`

### How to verify it
I've created some new tests in:
- test_contract.py
- test_interfaces.py

### Checklist

### DIFF
--- a/brownie/network/contract.py
+++ b/brownie/network/contract.py
@@ -217,6 +217,7 @@ class ContractContainer(_ContractBase):
         address: str,
         owner: Optional[AccountsType] = None,
         tx: Optional[TransactionReceiptType] = None,
+        persist: bool = True,
     ) -> "ProjectContract":
         """Returns a contract address.
 
@@ -245,7 +246,8 @@ class ContractContainer(_ContractBase):
         _add_contract(contract)
         self._contracts.append(contract)
         if CONFIG.network_type == "live":
-            _add_deployment(contract)
+            if persist:
+                _add_deployment(contract)
 
         return contract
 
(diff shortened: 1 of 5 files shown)


brownie/network/state.py
```diff
@@ -645,3 +645,32 @@ def _add_deployment(contract: Any, alias: Optional[str] = None) -> None:
 
     values = [contract._build.get(i) for i in DEPLOYMENT_KEYS]
     cur.insert(name, address, alias, all_sources, *values)
+
+
+def _remove_deployment(address: str, alias: str = None) -> Tuple[Optional[Dict], Optional[Dict]]:
+    if address and alias:
+        raise
```
banteg, line 652: raise a more specific error here

brownie/network/contract.py
```diff
@@ -1208,7 +1221,23 @@ def get_solc_version(cls, compiler_str: str, address: str) -> Version:
...
+        Arguments
+        ---------
+        address: str
+            An address to apply
+        alias: str | None
+            An alias to apply
+        """
+        return _remove_deployment(address, alias)
+
+    def set_alias(self, alias: Optional[str], persist: bool = True) -> None:
```
banteg, line 1240: check if alias can be used within the same process if not persisted

banteg (approved):

brownie/network/contract.py
mariuspod: it's still there as a field on the contract object regardless of the persist flag.

banteg (approved):

Issue #1281, opened 2021-10-07
ModuleNotFoundError: No module named 'Users.t'


bodasooqa:
This error shows up even if I just run a simple script:

<img width="527" alt="Снимок экрана 2021-10-08 в 01 27 10" src="https://user-images.githubusercontent.com/34233108/136470672-1ff82d6a-eef6-40e0-b5cf-0adf4f4a3a07.png">

```bash
t.lolaev@t-lolaev freaky-goblins % brownie run scripts/deploy.py 
Brownie v1.16.4 - Python development framework for Ethereum

FreakyGoblinsProject is the active project.

Launching 'ganache-cli --port 8545 --gasLimit 12000000 --accounts 10 --hardfork istanbul --mnemonic brownie'...
  File "brownie/_cli/run.py", line 49, in main
    return_value, frame = run(
  File "brownie/project/scripts.py", line 53, in run
    module = _import_from_path(script)
  File "brownie/project/scripts.py", line 149, in _import_from_path
    _import_cache[import_str] = importlib.import_module(import_str)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen, line line, in in
  File "<frozen, line line, in in
  ...
  File "<frozen, line line, in in
  File "<frozen, line line, in in
ModuleNotFoundError: No module named 'Users.t'
Terminating local RPC client...
```

Nyamgavaa:
**hi @bodasooqa**  are you know what happening my deployment. I deploy my contract. but it not deployed in rinkeby.etherscan, only making transaction.
I can not see my contract like mr.Patrick NFT/ERC-721/Collectible END-TO-END TUTORIAL | Deploy, List on Opensea, Host Metadata on IPFS lesson. https://www.youtube.com/watch?v=p36tXHX1JD8&t=126s **at 7.8**

bodasooqa:
@Nyamgavaa 

Hi!
If I were you, I would try to clone the project in its initial state, specify the necessary data and run the script
Also, make sure that you are looking specifically at the _Opensea_ **testnet** and that your _Metamask_ wallet is configured for the _Rinkeby_ network
So far there is too little input data that I could draw any conclusions :)

tuanhtr:
Thank you so much @bodasooqa, you saved my life!!!

Madgeniusblink:
> **UPD**
[...]

PR #1446, opened 2022-02-19
Update scripts.py

CountZer0:
errors on osx with:
File "<frozen, line line, in in

replaced _import_from_path()


freddie71010:
can you expand on what you actually did? your description of this change could use some more details.

CountZer0:
just copied some code from a stack overflow answer that imports modules better.

Might be a fix for this error: https://github.com/eth-brownie/brownie/issues/1281

freddie71010:
We shouldn't just arbitrarily be adding code in without testing it. Can you write a test for your change? or include the stack overflow question where you found the code from?

PR #1473, opened 2022-03-16
feat: allow cache disabling with forks

7xAquarius:
### Motivation
I keep getting `RPCProcessError: Unable to launch local RPC client.` when I try to fork Avalanche mainnet and testnet.
It's usually one of both working, while the other results in an error.

However, forking with `--fork.disableCache` on ganache fixes the issue.

### What I did
Added support for cache disabling with forks

### How I did it
Added support for the already existing `--fork.disableCache` option from ganache

### How to verify it
Add any forked network with `--disable_cache=true` and launch it
or in config file
```
networks:
    avax-fork:
        cmd_settings:
            disable_cache: true
```

### Sidenote

ganache also has the `--fork.deleteCache` option, both works for my issue but idk

### Checklist


WyseNynja:
I don't think we want to disable cache for all development networks. It can be very helpful on other chains. Can we change this to just disable for avalanche since it doesn't work there?

7xAquarius:
I agree.
This PR only allows the option to be used, it's not enabled by default on any network.

Maybe I should remove it from `test_validate_cmd_settings()`, is that what you meant ?

WyseNynja:
From your example comment, I just thought you'd set the default to "true". The default should be "false"

7xAquarius:
I've replaced `development` with `avax-fork` in my config file example, might be easier to understand you're right.

I don't think the default value needs to be set to `false` somewhere. Similar to the `fork` option, you set it when you need it.
In fact, it only works when `fork` option is set

iamdefinitelyahuman (approved):

Issue #1526, opened 2022-05-06
Add compatibility for anvil


chalex-eth:
Foundry just released anvil, which aim to replace ganache and hardhat RPC.
Foundry ecosystem is highly growing at speedlight, been able to run anvil in brownie console instead of ganache-cli would be awesome.

https://github.com/foundry-rs/foundry/tree/master/anvil

pratraut:
Need to add support for forking with block number. I think we need to add `fork_block` to a list at https://github.com/eth-brownie/brownie/blob/master/brownie/_cli/networks.py#L48

PR #1544, opened 2022-05-27
add anvil and docker build caching

WyseNynja:
### What I did

Related issue: #1526

### How I did it

Copy the binary of anvil from the official image into brownie. Currently, the "anvil" binary is only available in the "nightly" build. That should be in the "latest" (stable) image soon though.

### How to verify it

```
export DOCKER_BUILDKIT=1
docker-compose build
```

### Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,3 +6,5 @@ services:
     volumes:
       - .:/code
     command: tail -f /dev/null
+    # TODO: load this from an environment variable so M1 macs can easily use this
+    platform: linux/x86_64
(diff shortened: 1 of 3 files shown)


WyseNynja:
This also bumps the version of python in the docker image.

docker-compose.yml
```diff
@@ -5,4 +5,6 @@ services:
       context: .
     volumes:
       - .:/code
-    command: tail -f /dev/null
\ No newline at end of file
+    command: tail -f /dev/null
+    # TODO: load this from an environment variable so M1 macs can easily use this
+    platform: linux/x86_64
```
WyseNynja, line 10: I need to figure out the best way to do this. Without it, M1 macs will see `/lib64/ld-linux-x86-64.so.2: No such file or directory`

PR #1585, opened 2022-07-27
network: transaction: request memory content in traces

istankovic:
geth will not return memory content unless enableMemory is set to true,
which will cause _get_memory to fail.

This can be reproduced by provoking a typed error in a contract inside `brownie console` running with a local
geth-dev instance.

### Checklist


istankovic:
Could someone please rerun the CI? It seems like the tests are flaky.

czepluch:
Bump!

istankovic:
Another bump.

istankovic:
And another.

PR #1654, opened 2022-12-30
Update module import script

otis01:
Currently, running `brownie run scripts/deploy.py` fails when the path to the project has a `.` in it.

The following path
```
/Users/john.doe/Projects/Chainlink
```

Gets converted to
```
Users.john.doe.Projects.Chainlink
```
Which fails because it thinks that `john` and `doe` are two directories. Using the above method fixes this error.

### What I did

Add an import_module script that works regardless of the path.

### How I did it

Using "/" instead if "." in the path string.

### How to verify it

Run it locally in a path that has a `.` in it.

### Checklist

### DIFF
--- a/brownie/project/scripts.py
+++ b/brownie/project/scripts.py
@@ -147,13 +147,21 @@ def _get_path(path_str: str) -> Tuple[Path, Optional[Project]]:
     return path.resolve(), project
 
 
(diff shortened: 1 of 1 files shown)


BobTheBuidler:
@otis01 could you add some tests for this? I'd be happy to merge with those added

Issue #1661, opened 2023-01-20
Brownie is unable to verify contracts deployed other smart contracts


yzhuang:
### Environment information

* `brownie` Version: 1.19.1
* `solc` Version: 0.8.7
* Python Version: 3.10.5
* OS: osx

### What was wrong?
Contract.publish_source fails to verify contracts deployed by other smart contracts, which is used by smart contract deployments to vanity addresses.
No error message is given.

### How can it be fixed?
Brownie is not smart enough to look up the constructor parameters for contracts deployed by other contracts.
For example, see this contract deployed by an internal transaction: 
- https://etherscan.io/tx/0x07d10c02e14b8f468c0369b03f56a3ab229d0d116e1fa7f86b8781c7f33e550a
- Brownie fails to verify contracts deployed this way.

PR #1662, opened 2023-01-20
allow publish_source to take an additional constructor_args parameter

yzhuang:


yzhuang:
This PR is used to illustrate the problem for #1661, and this may not be the best fix.

PR #1683, opened 2023-04-22
Fix: Include explorer support for Optimism and Gnosis chain

Tritium-VLK:
### What I did
Just added a few lines to the regex to accept new env vars.  Note that the Optimistim scanner is optimistic.etherscan, so I put that one up top.

Also updated the starting network-config file to include mainnets and forks for the above chains.
Related issue: #

### How I did it
Just added a few lines that I am using locally.

### How to verify it
I'm a forked version actively to operate on these chains and run cross-chain reporting.

I did not see any tests for side chain explorers, so I did not add one.  I also could not find a fitting area to add more to the docs.

### Checklist


Tritium-VLK:
@skellet0r how does one get a review?  I'd love to move away from including my patch in dependency files.

charles-cooper (approved):
lgtm

PR #1703, opened 2023-07-30
Add base

Tritium-VLK:
## What I did
Just added a few lines to the regex to accept new env vars. Note that the Optimistim scanner is optimistic.etherscan, so I put that one up top.

Also updated the starting network-config file to include mainnets and forks for the above chains.

## How I did it
Just added a few lines that I am using locally.

## How to verify it
I'm a forked version actively to operate on these chains and run cross-chain reporting.

I did not see any tests for side chain explorers, so I did not add one. I also could not find a fitting area to add more to the docs.

## Checklist
 [x] I have confirmed that my PR passes all linting checks
 [x[ I have included test cases
 [x] I have updated the documentation
 [x] I have added an entry to the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Polygon zkEVM
 - Optimisim (fixed explorer support)
 - Gnosis Chain
+- Base chain
 
 ## [1.19.3](https://github.com/eth-brownie/brownie/tree/v1.19.3) - 2023-01-29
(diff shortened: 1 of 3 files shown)


Tritium-VLK:
@charles-cooper here's base chain also.  Then any chance for a new version number/tag so we can stop using a forked repo?

iamdefinitelyahuman:
Thanks for this.  Need to find a way to future-proof this, some standard format of `XXX_TOKEN` so new chains don't endlessly require PRs (especially since the maintainer is asleep at the wheel lol)

PR #1712, opened 2023-09-28
encoding='utf-8' for reading contracts

Hecate2:
### What I did
On Windows without utf-8 encoding by default, there can be errors like `UnicodeDecodeError: 'gbk' codec can't decode byte 0x80 in position 52: illegal multibyte sequence` when contracts are being read.
Related issue: None

### How I did it
Using encoding='utf-8' for contracts
### How to verify it
I cloned https://github.com/RockX-SG/stake and made brownie work with the contracts on Windows 10.
### Checklist


iamdefinitelyahuman:
Pretty straightforward fix, thanks

Issue #1360, opened 2021-12-03
Brownie cannot find return value.


0xrebased:
Steps to reproduce:

1) Deploy contract
2)Execute function call

![image](https://user-images.githubusercontent.com/93690142/144562235-6e8fd3d8-5724-42b1-b7db-cb21e0bc7dc7.png)

![image](https://user-images.githubusercontent.com/93690142/144562247-6d8137af-f393-41a6-ba3a-92bc08f1264d.png)

CALL STACK:

Looks like output is correct, but Brownie cannot find it for some reason. 

*{"output":"0000000000000000000000000000000000000000000000000000000000000006","gasUsed":"0x14a6","time":1107992}*

{"pc":0,"op":96,"gas":"0x1ef1","gasCost":"0x3","memory":"0x","memSize":0,"stack":[],"returnData":"0x","depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":2,"op":96,"gas":"0x1eee","gasCost":"0x3","memory":"0x","memSize":0,"stack":["0x80"],"returnData":"0x","depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":4,"op":82,"gas":"0x1eeb","gasCost":"0xc","memory":"0x","memSize":96,"stack":["0x80","0x40"],"returnData":"0x","depth":1,"refund":0,"opName":"MSTORE","error":""}
{"pc":5,"op":52,"gas":"0x1edf","gasCost":"0x2","memory":"0x","memSize":96,"stack":[],"returnData":"0x","depth":1,"refund":0,"opName":"CALLVALUE","error":""}
{"pc":6,"op":128,"gas":"0x1edd","gasCost":"0x3","memory":"0x","memSize":96,"stack":["0x0"],"returnData":"0x","depth":1,"refund":0,"opName":"DUP1","error":""}
{"pc":7,"op":21,"gas":"0x1eda","gasCost":"0x3","memory":"0x","memSize":96,"stack":["0x0","0x0"],"returnData":"0x","depth":1,"refund":0,"opName":"ISZERO","error":""}
{"pc":8,"op":96,"gas":"0x1ed7","gasCost":"0x3","memory":"0x","memSize":96,"stack":["0x0","0x1"],"returnData":"0x","depth":1,"refund":0,"opName":"PUSH1","error":""}
[...]

PR #1715, opened 2023-12-19
Include encoded return data in transaction trace

sparrowDom:
### What I did
The transaction trace didn't include the encoded return value of function calls if there was no source code contact information. This PR always includes the return data (when not null). If contract source code information is available, the `returndata` gets overridden by `return_value.` 

Related issue: https://github.com/eth-brownie/brownie/issues/1360

### How I did it
I've decoded the `returndata` in the loop that loops through the trace steps before it chooses to ignore a step that doesn't have contract source code information available.

### How to verify it
Run `call_trace(True)` on any transaction. Example before: 
Screenshot 2023-12-19 at 13 14 20
Example after (including this change): 
Screenshot 2023-12-19 at 13 14 33

### Checklist

### DIFF
--- a/brownie/network/transaction.py
+++ b/brownie/network/transaction.py
@@ -920,6 +920,20 @@ class TransactionReceipt:
                     last["address"], trace[i]["stack"][-2][-40:], trace[i]["stack"][-3]
                 )
 
+            # If the function signature is not available for decoding return data attach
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman:
Cool!

PR #1733, opened 2024-01-31
remove pythx

iamdefinitelyahuman:
### What I did
Remove `brownie analyze` and the `pythx` library as a dependency.  MythX appears to be deprecated in favor of a newer tool, and hard pins on `pythx` are getting in the way when trying to bump other dependency versions.


iamdefinitelyahuman:
Looks like a side-effect here is dropping support for python 3.7.  That seems fine.

iamdefinitelyahuman:
This ended up getting bigger than planned.

* remove pythx
* bump `pyyaml` (needed for 3.10)
* drop support for 3.7, test against 3.10

Things are still breaking with 3.11 onward, I think from `eth_abi`, but this seems like a logical place to merge

Issue #1757, opened 2024-03-27
Unable to determine error_code and test stops


YoshihitoAso:
### Environment information

* `brownie` Version: 1.20.2
* `ganache-cli` Version: 7.9.2
* `solc` Version: 0.8.23
* Python Version: 3.11
* OS: osx/linux

### What was wrong?

In my test, I use `brownie.reverts(revert_msg="Integer overflow")` to determine the error message, but this process stops with the current latest eth-brownie (v1.20.2) .
- Log: https://github.com/BoostryJP/ibet-SmartContract/actions/runs/8435203269/job/23101530020

After analyzing the cause, processing stopped at the following part.
- https://github.com/eth-brownie/brownie/blob/master/brownie/exceptions.py#L286

### How can it be fixed?

I haven't been able to investigate why it's been working fine so far, but I think this code should be implemented like this, for example:
```py
error_code = int(HexBytes(data[10:]).hex(), 16)
```

PR #1758, opened 2024-03-27
Fix error_code decoding

YoshihitoAso:
### What I did

Related issue: #1757

### How I did it

Please refer to diff.

### Checklist

### DIFF
--- a/brownie/exceptions.py
+++ b/brownie/exceptions.py
@@ -283,7 +283,7 @@ def decode_typed_error(data: str) -> str:
     selector = data[:10]
     if selector == "0x4e487b71":
         # special case, solidity compiler panics
-        error_code = int(data[4:].hex(), 16)
+        error_code = int(HexBytes(data[10:]).hex(), 16)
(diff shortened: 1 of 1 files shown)


brownie/exceptions.py
```diff
@@ -283,7 +283,7 @@ def decode_typed_error(data: str) -> str:
     selector = data[:10]
     if selector == "0x4e487b71":
         # special case, solidity compiler panics
-        error_code = int(data[4:].hex(), 16)
+        error_code = int(HexBytes(data[10:]).hex(), 16)
```
iamdefinitelyahuman, line 286: I think this is a safer approach, in some cases the returned data is already formatted to `HexBytes`.  Not sure why it's inconsistent to begin with..

```suggestion
        error_code = int(HexBytes(data)[4:].hex(), 16)
```

iamdefinitelyahuman (changes requested):

brownie/exceptions.py
iamdefinitelyahuman: ahh wait i didn't zoom out enough...  your approach is correct

iamdefinitelyahuman (approved):

PR #1774, opened 2024-05-10
fix: crash when waiting for a tx confirmation

banteg:
### What I did

in recent versions of anvil sometimes brownie ends up in a race condition when waiting for a tx confirmation. the bug results in`KeyError: 'blockNumber'`.

the root cause is anvil retuning no `blockNumber` field in `eth_getTransactionByHash` for pending transactions.

the bug was introduced here https://github.com/alloy-rs/alloy/pull/523

### How I did it

use `tx.get('blockNumber')` instead of `tx['blockNumber']`

### How to verify it

run a bunch of transactions, it doesn't happen every time. it's easier if you run anvil alongside, so you can see its output. the bug manifests when there is a `eth_getTransactionByHash` before `Transaction` printout.

```
eth_sendTransaction
eth_getTransactionByHash
eth_getCode

    Transaction: 0xc94c6c69570ca19763eda0a33a9b6a67803a5fa1fca2861ed2f4c08607edf0fb
    Gas used: 21000
```

### Checklist

### DIFF
--- a/brownie/network/transaction.py
+++ b/brownie/network/transaction.py
@@ -216,10 +216,10 @@ class TransactionReceipt:
         # await confirmation of tx in a separate thread which is blocking if
         # required_confs > 0 or tx has already confirmed (`blockNumber` != None)
         confirm_thread = threading.Thread(
-            target=self._await_confirmation, args=(tx["blockNumber"], required_confs), daemon=True
+            target=self._await_confirmation, args=(tx.get("blockNumber"), required_confs), daemon=True
         )
         confirm_thread.start()
-        if is_blocking and (required_confs > 0 or tx["blockNumber"]):
+        if is_blocking and (required_confs > 0 or tx.get("blockNumber")):
             confirm_thread.join()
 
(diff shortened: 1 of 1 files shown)


banteg:
fixed in https://github.com/alloy-rs/alloy/pull/719, so should be fixed in the next nightly anvil.

your call whether to merge this, technically brownie has followed the spec and anvil hasn't.

iamdefinitelyahuman:
Merging because this also fixes a related issue flagged by @michwill in an unrelated RPC client.  Not sure the root cause but this solves, and it's easier to handle here than try to fix somewhere upstream.

lemenkov:
Funny but [this field is required](https://github.com/ethereum/execution-apis/blob/c4e7da7e486f6cb6ff6281c3ff28bb37445d5415/src/schemas/transaction.yaml#L334-L339) and yet something weird happened recently.

banteg:
the root cause was this https://github.com/alloy-rs/alloy/pull/719

PR #1784, opened 2024-06-26
fix: anvil mine

dudesahn:
### What I did

small refactor so `chain.mine()` works as expected w/ anvil

Currently, `chain.mine()` or `chain.mine(x)` both revert with error `brownie.exceptions.RPCRequestError: Timestamp error: 1 is lower than or equal to previous block's timestamp`

Related issue: #

### How I did it

set the `mine` function to only use `evm_mine` and to pass `blocks` instead of defaulting to hardcoded 1

### How to verify it

- `chain.mine()` mines 1 block
- `chain.mine(x)` mines x blocks

### Checklist

### DIFF
--- a/brownie/network/rpc/anvil.py
+++ b/brownie/network/rpc/anvil.py
@@ -75,10 +75,8 @@ def sleep(seconds: int) -> int:
     return seconds
 
 
-def mine(timestamp: Optional[int] = None) -> None:
-    if timestamp:
-        _request("evm_setNextBlockTimestamp", [timestamp])
-    _request("evm_mine", [1])
+def mine(blocks: Optional[int] = None) -> None:
+    _request("evm_mine", blocks)
 
 
 def snapshot() -> int:


docs/api-network.rst
```diff
@@ -2184,13 +2184,11 @@ Chain Methods
         >>> chain.time()
         1550189143
 
-.. py:method:: Chain.mine(blocks=1, timestamp=None, timedelta=None)
```
iamdefinitelyahuman, line 2187: `timestamp` and `timedelta` should still work when using ganache / hardhat so removing the documentation around it feels wrong.

iamdefinitelyahuman (changes requested):
The fix is good, please adjust the docs edit and then happy to merge and release it.

docs/api-network.rst
dudesahn: updated!

BobTheBuidler (approved):

PR #1814, opened 2025-03-21
chore: refactor events with `map`

BobTheBuidler:
### What I did

This PR will eliminate 1 list comprehension and x-2 global name lookups, where x is the number of logs in the iterable, per _decode_logs call

Might seem silly but profiling showed we spent an inordinate amount of time on these lines.

Related issue: #

### Checklist


Mikke84 (approved):

Mikke84 (changes requested):
@mikaeljfjf.8194

PR #1839, opened 2025-04-22
feat(test): run tests verbosely

BobTheBuidler:
### What I did
One of the test failures in the runner shows:
```
    def test_getitem_negative_index(devnetwork, accounts, chain, web3):
        block = chain[-1]
>       assert block == web3.eth.get_block("latest")
E       AssertionError: assert AttributeDict...'uncles': []}) == AttributeDict...drawals': []})
E         Use -v to get the full diff
```

so I decided to modify the test runner so the tests fail verbosely
Related issue: #

### How I did it
added the -v flag to each test command in tox.ini

### How to verify it
the tests will run on this PR

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -21,13 +21,13 @@ deps =
     docs-{local,external}: sphinx_rtd_theme
     docs-{local,external}: pygments_lexer_solidity
 commands =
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman (approved):
-v

PR #1840, opened 2025-04-22
fix(test): add handler for race condition in `network` fixture setup

BobTheBuidler:
Maybe a better solution here would be a threading.Lock?


BobTheBuidler:
threading.Lock implemented

PR #1857, opened 2025-05-16
fix: _to_wei regression from breaking change in hexbytes

BobTheBuidler:
### What I did

hexbytes v1 broke HexBytes.hex, meaning we need to make some changes for compatability

This PR makes brownie work on either version, and we might want to setup a testing workflow for both <1 and >1 to ensure we're good to go. Thoughts?

Related issue: #

### Checklist


brownie/convert/datatypes.py
```diff
@@ -98,6 +98,8 @@ def _to_wei(value: WeiInputTypes) -> int:
     original = value
     if isinstance(value, bytes):
         value = HexBytes(value).hex()
```
BobTheBuidler, line 100: why does this need to be converted to a HexBytes? I think we can just do 

value = value.hex()
if value:
    return int(value, 16)

brownie/convert/main.py
```diff
@@ -63,12 +63,12 @@ def to_bool(value: Any) -> bool:
...
 def to_string(value: Any) -> str:
     """Convert a value to a string"""
-    if isinstance(value, bytes):
-        value = HexBytes(value).hex()
-    value = str(value)
-    if value.startswith("0x") and eth_utils.is_hex(value):
-        try:
+    try:
+        if isinstance(value, bytes):
+            return eth_utils.to_text(hexstr=HexBytes(value).hex())
```
BobTheBuidler, line 68: I think we can just value.hex() here too

BobTheBuidler:
This also fixes a breaking change in eth-account as seen [here](https://eth-account.readthedocs.io/en/latest/release_notes.html#removals)

brownie/convert/datatypes.py
iamdefinitelyahuman: been a while, can't say i remember.  i think it had to do with 0-length hex values:

```py
# output for version <1
>>> HexBytes('0x').hex()
"0x"

# output for version >=1
>>> HexBytes('0x').hex()
""
```

i know this inconsistency was the source of a lot of headaches, so i tended to aggressively convert bytes -> hexbytes in order to have consistent behavior on `.hex()`

brownie/network/account.py
```diff
@@ -939,13 +946,22 @@ def sign_defunct_message(self, message: str) -> SignedMessage:
...
+        if ETH_ACCOUNT_LT_0_13_0:
+            return SignedMessage(
+                messageHash=msg_hash_bytes,
+                r=r,
+                s=s,
+                v=v,
+                signature=HexBytes(eth_signature_bytes),
+            )
+        else:
+            return SignedMessage(
```
iamdefinitelyahuman, line 958: facepalm for upstream breaking changes lol.

is there a reason you need to support multiple versions of `eth-account`?

brownie/network/contract.py
```diff
@@ -130,7 +131,11 @@ def decode_input(self, calldata: Union[str, bytes]) -> Tuple[str, Any]:
         if not isinstance(calldata, HexBytes):
             calldata = HexBytes(calldata)
 
-        fn_selector = calldata[:4].hex()  # type: ignore
+        if HEXBYTES_LT_1_0_0:
```
iamdefinitelyahuman, line 134: must be a cleaner way here

```py
fn_selector = calldata[:4].hex()
if not fn_selector.startswith('0x'):
    fn_selector = f"0x{fn_selector}"
```

tho this is where you probably tell me that your method is a few milliseconds faster :rofl: 
```

Comment (iamdefinitelyahuman) [reply brownie/network/contract.py:134]: Ahh I see you're using `removeprefix` everywhere

```py
fn_selector = f"0x{calldata[:4].hex().removeprefix('0x')"
```

is that too many things in one line?  seems to address everything without having to import from another file just to verify a package version number.

brownie/utils/hex.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def hexstring(value: bytes) -> HexStr:
+        """Convert a bytes value to a hexstring on hexbytes<1."""
+        return HexBytes(value).hex()
+
+else:
+
+    def hexstring(value: bytes) -> HexStr:
+        """Convert a bytes value to a hexstring on hexbytes>=1."""
+        return f"0x{HexBytes(value).hex()}"
```
iamdefinitelyahuman, line 29: same as before (and a lot of other places we could surely apply this, unless you have a specific reason for doing it the way you are)

```python
def hexstring(value: bytes) -> HexStr:
    return f"0x{HexBytes(value).hex().removeprefix('0x')}"
```

think we could get rid of this ugly `HEXBYTES_LT_1_0_0` altogether

iamdefinitelyahuman (commented):
overall it's cool and if you can fight me on my comments successfully, we can merge today

brownie/utils/hex.py
BobTheBuidler: was just for speed since my server is running thru these functions 24/7

we can probably get rid of the constant in the other files while keeping the optimized def in this file, wdyt?

brownie/network/account.py
BobTheBuidler: yeah, it was just a dep conflict. if we're going to support multiple hexbytes we need to support multiple eth-account too because anybody pinned to hexbytes<1 can't use the current eth-account

I figured since we release eth-event in a way that serves those people, we would do the same in here

brownie/network/transaction.py
```diff
@@ -568,7 +572,11 @@ def _set_from_tx(self, tx: Dict) -> None:
             contract = state._find_contract(tx.get("to"))
             if contract is not None:
                 self.contract_name = contract._name
-                self.fn_name = contract.get_method(tx["input"].hex())
+                if HEXBYTES_LT_1_0_0:
```
BobTheBuidler, line 575: what if I just make a 2nd util in utils.hex that covers this case, and we just replace these instances with that helper to make everything cleaner?

brownie/utils/hex.py
BobTheBuidler: wait I think if I just take out the call to HexBytes here, and just build the hexstring from bytes.hex, we can use this util in all the places we currently have HEXBYTES_LT_1_0_0

That's my new plan, I'll implement if you agree

iamdefinitelyahuman (approved):

PR #1873, opened 2025-05-25
fix(test): include setuptools in test deps to fix web3py

BobTheBuidler:
### What I did

Related issue: #

### Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,13 +6,13 @@
 #
 aiohappyeyeballs==2.6.1
     # via aiohttp
-aiohttp==3.12.0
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -264,3 +264,6 @@ yarl==1.20.0
     # via aiohttp
 zipp==3.21.0
     # via importlib-metadata
+
+# The following packages are considered to be unsafe in a requirements file:
+# setuptools
```
BobTheBuidler, line 269: wtf is this, we need to install setuptools

PR #1974, opened 2025-08-09
docs(convert): Fix typo in Wei docstring example

gap-editor:
The docstring for the `Wei` class contained a hex string example with
a stray backslash at the end.

This commit removes the unnecessary character to correct the example
and prevent confusion for developers.


BobTheBuidler (approved):

BobTheBuidler:
Good catch, thanks!

PR #2246, opened 2026-06-04
Update README to remove maintenance notice

0xInnominatus:
Removed the note about Brownie no longer being actively maintained

### What I did

Considering the frequency of releases and the active use of eth-brownie in many projects, including Yearn, maybe it makes sense to remove the mention that the framework is no longer actively maintained from docs and README? I personally use eth-brownie together with eth-wake quite actively and I am very happy to see that the good old tool is back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@
 
 Brownie is a Python-based development and testing framework for smart contracts targeting the [Ethereum Virtual Machine](https://solidity.readthedocs.io/en/v0.6.0/introduction-to-smart-contracts.html#the-ethereum-virtual-machine).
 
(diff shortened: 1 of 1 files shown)


0xInnominatus:
@BobTheBuidler Hi Bob! Could you please share your thoughts on this topic? I’m planning to start a new project in the near future, and if Brownie is actively maintained again, I’d like to use this framework, especially since you’ve added mypyc support and I critically need high performance (Ape and Wake currently don’t support running with mypyc). What do you think? Can we expect the disclaimer about the lack of active maintenance to be removed from the README and documentation?

PR #2265, opened 2026-08-31
test: write invalid coverage fixture without leaking a handle

ginavalent:
### What I did

Changed the invalid coverage-report fixture to use `Path.write_text(..., encoding="utf-8")` instead of calling `write` on an unclosed file object.

Related issue: N/A

### How I did it

The pathlib convenience method opens and closes the file within the call, making ownership explicit while preserving the exact fixture contents.

### How to verify it

Run `pytest tests/test/plugin/test_output.py -k coverage_save_report`.

Local verification performed:

- `python3 -m py_compile tests/test/plugin/test_output.py`
- `git diff --check`

The focused pytest test was not run locally because pytest is not installed in this environment.

### Checklist

- [ ] I have included test cases — existing test adjusted; no new behavior introduced
- [ ] I have updated the documentation — not applicable
- [ ] I have added an entry to the changelog — not applicable to a test-only cleanup


ginavalent:
@BobTheBuidler Could you please take a look when you have a chance? Thanks!

